PR #14, opened 2015-12-09
Remove destructor

aradi:
Let's disable the destructor in the type `bml_matrix_t` again ( :-1: ), requiring the users to use explicit deallocation routines in order to avoid memory leaks, but restoring compatibility with the buggy gfortran compiler. ( :+1: ).

### DIFF
--- a/src/Fortran-interface/bml_allocate_m.F90
+++ b/src/Fortran-interface/bml_allocate_m.F90
@@ -30,8 +30,9 @@ contains
     character(len=*), intent(in) :: matrix_type, element_type
     integer, intent(in) :: element_precision
     integer(C_INT), intent(in) :: n, m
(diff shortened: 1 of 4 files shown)


aradi:
Apparently, on travis, various Fortran tests for `bml_scale()` fail. Interestingly, when looking on the raw-output, the matrices printed out are identical. Nevertheless, sometimes the test system reports completely crazy differences while in other cases it reports a difference below tolerance, and let the test fail nevertheless. I think, this would be worth of being investigated.

nicolasbock:
Thanks. I'll have a look.

nicolasbock:
@aradi I opened an issue regarding the failures of bml_scale(). -- Issue #15

PR #22, opened 2017-02-09
Updated README and Travis-CI configuration

nicolasbock:
We require >=OpenMP-3.1 which was added to gcc-4.7. I made a note of
that in the README and removed the Travis compiler tests for <gcc-4.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
-language: c
+# We build on a matrix across difference compiler versions. If we specify C or
+# C++ for the language, then Travis-CI will export either the CC or CXX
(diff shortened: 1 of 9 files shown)


nicolasbock:
@cnegre Still working on the tests. There are some memory leaks we need to address before this PR will pass the Travis-CI testsuite. I'll update this PR soon.

nicolasbock:
@suemni Could you have a look at the Travis-CI test failures? The `default_domain` is not properly de-allocated and leads to a memory leak. I think the problem might be that the `bml_copy` does not copy the domain which leads to issues upon deallocation.

PR #24, opened 2017-02-10
Updated README

nicolasbock:
I removed the lanl internal email list. This list is not accessible by
the public and as such useless for discussions of an open sourced code.


cnegre:
Merged?

PR #54, opened 2017-04-07
Remove unnecessary assert

nicolasbock:
The assert in the comparison function is not necessary. It is not
illegal to have to equal elements.


nicolasbock:
Can we have a sanity check such as "we shouldn't have two equal indices in that array" to somewhere else? The `qsort` function specifies what the comparison function should return and that includes `0` for equal elements.

PR #58, opened 2017-04-27
added diag from ellpack to ellpack

nicolasbock:


nicolasbock:
@cnegre Could you have a look at the failed tests? Maybe there are some typing problems in the test driver?

PR #67, opened 2017-06-13
Added bml_inverse

suemni:
Added code for bml_inverse for dense and ellpack.
Also added test.

I have changed the bml_inverse API to B = bml_inverse(A)


src/C-interface/dense/bml_inverse_dense_typed.c
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * \param A Matrix A
+ * \param B Inverse of Matrix A
+ */
+void TYPED_FUNC(
+    bml_inverse_dense) (
+    const bml_matrix_dense_t * A,
+    bml_matrix_dense_t * B)
+{
+    memcpy(B->matrix, A->matrix, A->N * A->N * sizeof(REAL_T));
```
nicolasbock: Without a check whether `B` has the right size (i.e. matches the size of `A`) this is not safe.

src/C-interface/ellpack/bml_inverse_ellpack_typed.c
```diff
@@ -0,0 +1,68 @@
...
+    bml_free_memory(A_dense);
+
+    // Calculate inverse in dense_bml
+    TYPED_FUNC(bml_inverse_inplace_dense) (D);
+
+    A_dense = bml_convert_to_dense_dense(D, dense_row_major);
+    bml_deallocate_dense(D);
+
+    // From dense_array to ellpack_bml
+    B = bml_convert_from_dense_ellpack(A->matrix_precision,
```
nicolasbock: This assignment will get lost on function exit and causes a memory leak.

src/Fortran-interface/bml_c_interface_m.F90
```diff
@@ -343,6 +343,12 @@ function bml_identity_matrix_C(matrix_type, matrix_precision, n, m, &
       type(C_PTR) :: bml_identity_matrix_C
     end function bml_identity_matrix_C
 
+    subroutine bml_inverse_C(a, b) bind(C, name="bml_inverse")
+      import :: C_PTR
+      type(C_PTR), value, intent(in) :: a
+      type(C_PTR), value, intent(in) :: b
```
nicolasbock: The intent of `b` should be `inout`.

src/C-interface/bml_inverse.c
```diff
@@ -0,0 +1,30 @@
...
+#include "bml_utilities.h"
+#include "bml_logger.h"
+#include "bml_types.h"
+#include "dense/bml_inverse_dense.h"
+#include "ellpack/bml_inverse_ellpack.h"
+
+void
+bml_inverse(
+    const bml_matrix_t * A,
+    bml_matrix_t * B)
```
nicolasbock: I think it would be clearer if the interface was `bml_matrix_t * bml_inverse(const bml_matrix_t * A)` to indicate that a new matrix is calculated and allocated.

src/C-interface/ellpack/bml_inverse_ellpack_typed.c
suemni: I have changed the bml_inverse API to B = bml_inverse(A) as suggested.

tests/bml_test.c
```diff
@@ -12,7 +12,8 @@ const int NUM_TESTS = 18;
 
 const char *test_name[] = {
     "add", "allocate", "adjacency", "adjungate_triangle", "convert", "copy",
-    "diagonalize", "inverse", "multiply", "norm", "normalize", "scale", "set_diagonal",
+    "diagonalize", "inverse", "multiply", "norm", "normalize", "scale",
+        "set_diagonal",
```
nicolasbock: Could you move this string into the next line so that it's not by itself? Thanks!

suemni:
I have no idea what that means.

Please just approve the new bml_inverse, so I can get on with my work.


- Sue

PR #92, opened 2017-07-29
Make Travis-CI check indentation

nicolasbock:
The indentation of the .c and .h files should be verified by Travis-CI.
We used the command `indent` which simply indents all files instead of
the command `check_indent` which verifies that the indentation of those
files is correct.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/92)


nicolasbock:
@cnegre We need to merge #91 first before this will pass Travis-CI.

nicolasbock:
Ok, @cnegre this one is good to go.

PR #95, opened 2017-08-01
bml_import does not copy M

nicolasbock:
The code added in eb87291d7bcbc0b6c61b0a6f41ec22e07a92f025 is supposed
to insist on an M value for non-dense matrices and replace the default
of `m = 0` with `m = n`. However, due to incorrect logic this change
ended up setting `m` only for dense matrices and ignoring it otherwise.
The resulting bml matrix was incorrect.


nicolasbock:
Thanks!

PR #96, opened 2017-08-01
Add missing bml_scale Fortran interfaces

nicolasbock:
With change b93a366260afdd0ac0b0be4a5970e51173c9b7ed type matching scale
factors where introduced to the `bml_scale` APIs. However, only the
`bml_scale_two` Fortran interface was adapted but not the
`bml_scale_one` interfaces. This change adds the necessary APIs.


nicolasbock:
Thanks!

Issue #102, opened 2017-08-16
Compilation error with gcc 4.6.1


cnegre:
Compilation error with gcc 4.6.1 

~~~
[  0%] Building C object src/C-interface/dense/CMakeFiles/bml-dense-single_real.dir/bml_normalize_dense_typed.c.o
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c: In function ‘bml_gershgorin_dense_single_real’:
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:76: error: expected ‘+’, ‘*’, ‘-’, ‘&’, ‘^’, ‘|’, ‘&&’, or ‘||’ before ‘max’
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:77: error: expected ‘+’, ‘*’, ‘-’, ‘&’, ‘^’, ‘|’, ‘&&’, or ‘||’ before ‘min’
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:101: error: ‘emax’ not specified in enclosing parallel
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:71: error: enclosing parallel
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:103: error: ‘emin’ not specified in enclosing parallel
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:71: error: enclosing parallel
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c: In function ‘bml_gershgorin_partial_dense_single_real’:
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:149: error: expected ‘+’, ‘*’, ‘-’, ‘&’, ‘^’, ‘|’, ‘&&’, or ‘||’ before ‘max’
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:150: error: expected ‘+’, ‘*’, ‘-’, ‘&’, ‘^’, ‘|’, ‘&&’, or ‘||’ before ‘min’
/turquoise/users/cnegre/bml/src/C-interface/dense/bml_normalize_dense_typed.c:165: error: ‘emax’ not specified in enclosing parallel
~~~

nicolasbock:
[...]

PR #103, opened 2017-08-17
Replace min/max reductions for <gcc-4.7 support

nicolasbock:
The min/max reductions in C/C++ were added to the OpenMP standard in
OpenMP 3.1 which is available in >=gcc-4.7. In order to support earlier
compilers we have to replace all occurrences of min/max reduction
clauses with an appropriate alternative.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/103)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,17 +18,20 @@ addons:
             - cmake
             - cmake-data
             - doxygen
+            - g++-4.6
             - g++-4.7
             - g++-4.8
             - g++-4.9
             - g++-5
             - g++-6
+            - gcc-4.6
             - gcc-4.7
             - gcc-4.8
             - gcc-4.9
             - gcc-5
             - gcc-6
             - gfortran
+            - gfortran-4.6
             - gfortran-4.7
             - gfortran-4.8
             - gfortran-4.9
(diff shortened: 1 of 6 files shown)


nicolasbock:
The impact on performance depends strongly on where the critical section is. In this particular case however, I am not convinced that it will have a large impact. The critical section is after the parallel loop and serializes the min/max finding. The overhead incurred by the mutex is presumably very small. 

An alternative implementation would use a shared array of min/max variables and then find the global min/max in the master thread. However, this approach has the drawback that it requires a barrier after the loop (which we avoid now with the `nowait` clause) which also impacts performance.

And lastly, using ifdefs will make the code a lot less readable and I am reluctant to go down that route unless it becomes necessary.

Before we jump into any premature optimizations, we should benchmark the code to find out what performance impact we get with this PR.

nicolasbock:
@suemni I have pushed #105 with the aforementioned alternative implementation. I have not done any performance tests.

nicolasbock:
@suemni Could you re-review this? @cnegre is blocked on this issue (see #102).

nicolasbock:
@suemni I have updated the example code and ran some performance comparisons. As far as I can tell there is not difference between the two versions.

suemni:
So why does BML have to support <gcc4.6?

I would like to be assured that my routines are running with OpenMP w reductions.

We should not be coding to old compilers needs,

nicolasbock:
We should be coding towards the needs of the users of this library. Issues #102 is currently open and there is a use case for that compiler. The change is pretty small and I don't see how we are now "coding to old compilers". We are not using any other feature that is not supported in gcc-4.6.

suemni:
So who is the user that made that request?

Our main customer is ECP and they have not requested that.

I want to be able to use OpenMP with reductions. This is the default.

Please use ifdefs for <gcc4.7 code.


- Sue

nicolasbock:
The use of ifdefs should be avoided in my opinion because we would have to maintain two almost identical code paths increasing effort on our end for no other reason than to be able to use the 'min/max' reduction in three places in the code. As I stated earlier, #102 is still open, i.e. there is a user request for this feature. Finally, I don't see the bml's sole purpose as to feed into the ECP. It is a generally usable library and should be usable as such.

suemni:
So again, who is the user that requested gcc4.6? And have they provided funding to support this?

We do not have the resources to handle this.

nicolasbock:
Funding is a non-issue since this project is a community effort. Unless of course you want to pay me for my time on this project, then I might be swayed to reconsider this stance :wink:.

nicolasbock:
Closing.

PR #108, opened 2017-08-21
Use appropriate values for initial min/max

nicolasbock:
Using a finite value to initialize min/max variables has the
disadvantage that any value might be too small or too large. Instead, we
should use the type dependent representable min/max value.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/108)


nicolasbock:
Ping...

nicolasbock:
Thanks!

Issue #90, opened 2017-07-28
Bml to bml converter


cnegre:
We should have a bml to bml converter where
The idea is to change precitions and types as needed.

nicolasbock:
Are you thinking of something like `single_real` to `complex_double` for example?

cnegre:
Yes, that's the kind of thing we might need

nicolasbock:
Closed via #113

PR #113, opened 2017-09-01
Split import/export into separate modules

nicolasbock:
This change splits the import/export Fortran APIs into two separate
modules. This change mirrors the previous changes to the C API and
prepares the implementation of a true conversion API (See Issue #90).

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/113)


tests/convert_matrix_typed.c
```diff
@@ -0,0 +1,43 @@
...
+    bml_matrix_t *A;
+    REAL_T *A_dense;
+    REAL_T *B_dense;
+
+    A_dense = bml_allocate_memory(sizeof(REAL_T) * N * N);
+    for (int i = 0; i < N * N; i++)
+    {
+        A_dense[i] = rand() / (double) RAND_MAX;
+    }
+    A = bml_import_from_dense(matrix_type, matrix_precision, dense_row_major,
```
nicolasbock: @cnegre How should we test the `bml_convert` function? We could generate a dense matrix, convert that to bml, and then convert that bml to another bml. What do you think?

The other question we should think about is the following: Since `bml_convert` can convert from any bml matrix to any bml matrix, we should really run `(N_t N_p)^2` tests, where `N_t` is the number of matrix types and `N_p` the number of matrix precisions. Currently we can't do that automatically through CMake. Do we also want to add some mechanism in CMake to allow for this double testing? What do you think?

PR #142, opened 2017-09-27
Explicitly Including mpi.h

rspavel:
Explicitly including mpi.h in bml_parallel.h due to issues with spack
build and dependencies

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/142)

### DIFF
--- a/src/C-interface/bml_parallel.h
+++ b/src/C-interface/bml_parallel.h
@@ -14,6 +14,7 @@
 #endif
 
 #ifdef DO_MPI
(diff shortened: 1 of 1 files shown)


src/C-interface/bml_parallel.h
```diff
@@ -14,6 +14,7 @@
 #endif
 
 #ifdef DO_MPI
+#include "mpi.h"
```
nicolasbock: Please use `<mpi.h>`.

Issue #144, opened 2017-10-03
Compilation without omp


cnegre:
When BML_OMP is set to no the compilation fails cause omp.h is still included in 
the source code and not protected with "ifdefs"

suve:
Hi. I took a quick look at the code (`master` branch) and the only place I've found where `#include <omp.h>` isn't guarded by IFDEF is `src/C-interface/bml_allocate.c`. In all the other files the include is guarded, albeit by the `_OPENMP` symbol, not `BML_OMP`. (By the way, did you mean `BML_OPENMP`? The shorter `BML_OMP` doesn't appear anywhere within the project.)

nicolasbock:
Hi @suve, thanks for looking into this! You are right, that's the only place I found as well. It would be great if you could send us a pull request with a fix!

To explain further, in https://github.com/lanl/bml/blob/master/CMakeLists.txt#L113 you'll find that we use `BML_OPENMP` as a switch to `CMake` to indicate that the user would like to enable `OpenMP`. If `CMake` finds a working `OpenMP` implementation, in particular, if `>= OpenMP-3.1`, then `OpenMP` is enabled causing the compiler to define `_OPENMP` which we use to guard the `#include <omp.h>`. Of course if the compiler does not support `OpenMP` then the code should not `#include <omp.h>`.

Thanks!

nicolasbock:
This is closed with #149

PR #149, opened 2017-10-18
Add an include-guard for <omp.h> in bml_allocate.c

suve:
Does what is says ~on the tin~ in the title. Should fix issue #144.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/149)

### DIFF
--- a/src/C-interface/bml_allocate.c
+++ b/src/C-interface/bml_allocate.c
@@ -74,7 +74,11 @@ bml_allocate_memory(
     else
     {
         void *ptr = malloc(size);
+#ifdef _OPENMP
         int nt = omp_get_num_threads();
+#else
+        int nt = 1;
+#endif
         int step = size / nt;
         int maxi = step * (nt - 1);
         int r = size - maxi;
--- a/src/C-interface/bml_allocate.c
+++ b/src/C-interface/bml_allocate.c
@@ -12,7 +12,9 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef _OPENMP
 #include <omp.h>
+#endif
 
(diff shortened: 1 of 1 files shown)


src/C-interface/bml_allocate.c
```diff
@@ -12,7 +12,9 @@
 #include <stdlib.h>
 #include <string.h>
 
+#ifdef _OPENMP
 #include <omp.h>
+#endif
```
nicolasbock: Thanks!

It looks like as though this change broke the call at: https://github.com/lanl/bml/blob/master/src/C-interface/bml_allocate.c#L75. Could you `#ifdef` that call as well and set `nt = 1` in case the code is not built with OpenMP?

Thanks again for the PR!

suve: Oh, right. I didn't actually built it and I forgot to `grep` for function calls. It seems that was the only `omp_*` function call not guarded by an `#ifdef`. Fixed.

nicolasbock:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please limit commit subject line to 50 characters.<br />
0f3a3ab1182c1643a3c77566ff38d1706c89af0d</td>
    </tr>
  </tbody>
</table>

  Generated by :no_entry_sign: <a href="http://danger.systems/">Danger</a>
</p>

nicolasbock:
Thanks!

PR #155, opened 2017-10-27
Changed names

cnegre:
For some reason in this way I'm able to set the flags correctly. With the names before 
it was not able to get the "last" flag of the string I was passing.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/155)


cnegre:
This works correctly:
     $ CMAKE_C_FLAGS="-ggg" ./build.sh configure
This does not: 
     $ CMAKE_C_FLAGS="-ggg -O2" ./build.sh configure

Also, the *FLAG name is more consistent with EXTRA_*FLAGS

nicolasbock:
I can confirm that. It's an issue with string quoting in Bash. I am working on a fix.

cnegre:
PR no needed anymore ....

PR #172, opened 2017-12-06
Enable CI testing on OS X

nicolasbock:
We keep running into issues only seen on OS X systems. Since Travis-CI
offers OS X based tests, we can leverage it to fix our OS X build
issues. This change enables testing on OS X.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/172)

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -45,6 +45,39 @@ jobs:
       - name: Build docs
         run: ./build.sh --debug docs
 
+  build_osx:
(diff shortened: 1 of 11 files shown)


suemni:
Sounds good, does it fix the problem recently reported?

nicolasbock:
I am waiting for Travis-CI to report back and hopefully reproduce the problem. If that's the case I will add a commit to fix the issue. I don't have direct access to a system running OS X, which is why I have to go about it a bit indirectly by iterating through CI.

PR #190, opened 2018-03-11
Test for multiply_x2

cnegre:
Test for bml_multiply_x2

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/190)

### DIFF
--- a/tests/multiply_matrix_x2_typed.c
+++ b/tests/multiply_matrix_x2_typed.c
@@ -85,7 +85,7 @@ int TYPED_FUNC(
     const double alpha = 1.0;
     const double beta = 0.0;
     const double threshold = 0.0;
-    REAL_T* trace;
+    REAL_T *trace;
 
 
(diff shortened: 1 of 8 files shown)


cnegre:
@nicolasbock I don't understand the CI error that we get on the 907.1 build. Any clues 
of what this can be?

tests/multiply_matrix_x2_typed.c
```diff
@@ -85,7 +85,8 @@ int TYPED_FUNC(
     const double alpha = 1.0;
     const double beta = 0.0;
     const double threshold = 0.0;
-    const double trace;
+    REAL_T* trace;
```
nicolasbock: @cnegre The indentation here is not correct. Have a look at the Travis-CI error message and the diff towards the end. But basically you need to move the `*` so that it reads `REAL_T *trace;`.

PR #243, opened 2019-04-08
Bux fix in memory allocation for MAGMA matrices

jeanlucf22:
MAGMA matrices memory allocation was not set to 0 in some cases, which was not consistent with non-MAGMA based dense format. It also lead to test failures on Summit@OLCF.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/243)

### DIFF
--- a/src/C-interface/dense/bml_allocate_dense_typed.c
+++ b/src/C-interface/dense/bml_allocate_dense_typed.c
@@ -73,6 +73,8 @@ bml_matrix_dense_t *TYPED_FUNC(
     magma_int_t ret = MAGMA(malloc) ((MAGMA_T **) & A->matrix,
                                      A->ld * matrix_dimension.N_rows);
     assert(ret == MAGMA_SUCCESS);
+
+    bml_clear_dense(A);
 #else
     A->ld = matrix_dimension.N_rows;
     A->matrix =


src/C-interface/dense/bml_allocate_dense_typed.c
```diff
@@ -73,6 +73,8 @@ bml_matrix_dense_t *TYPED_FUNC(
     magma_int_t ret = MAGMA(malloc) ((MAGMA_T **) & A->matrix,
                                      A->ld * matrix_dimension.N_rows);
     assert(ret == MAGMA_SUCCESS);
+
+    bml_clear_dense(A);
```
nicolasbock: Why is this call needed? Does `MAGMA(malloc)` write something into `A`?

nicolasbock: Oh, ok, now I see that you explain what you are doing in the PR description. Thanks! Could you copy that explanation into the commit message? The PR description is only visible on GitHub and doesn't help someone trying to understand the code using only git without GitHub.

Thanks!

nicolasbock: No, it's unlimited. The common convention though is to structure it like an email. First line is the subject, and that's typically limited to something like 50 characters, then an empty line and then the message body, line wrapped to about 72 characters, as many lines as you like. You can find a nice description here: https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html.

jeanlucf22: I just amended my commit and added my comments into the commit message

PR #244, opened 2019-04-15
Add new format 'ellblock'

jeanlucf22:
This is based on the ellpack format, but each "element" is now a block.
Block sizes are defined either by the first call to build a matrix with
'bml_block_matrix' or by their default values. They have to be the same
for all matrices. This code seems to work properly and pass all the tests.
It has however not been optimized and does not include OpenMP directives.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/244)

### DIFF
--- a/src/C-interface/CMakeLists.txt
+++ b/src/C-interface/CMakeLists.txt
@@ -85,4 +85,5 @@ endif()
 
 add_subdirectory(dense)
 add_subdirectory(ellpack)
+add_subdirectory(ellblock)
 add_subdirectory(ellsort)
--- a/src/C-interface/bml_add.c
+++ b/src/C-interface/bml_add.c
@@ -4,6 +4,7 @@
 #include "dense/bml_add_dense.h"
 #include "ellpack/bml_add_ellpack.h"
 #include "ellsort/bml_add_ellsort.h"
+#include "ellblock/bml_add_ellblock.h"
 
 #include <stdlib.h>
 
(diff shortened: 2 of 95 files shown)


jeanlucf22:
There seems to be an issue with the indent script when applied to the file src/C-interface/ellblock/bml_transpose_ellblock_typed.c. At every application, it changes something, back and forth...

jeanlucf22:
I changed the code to avoid the issue. Now there seems to be another issue I don't understand.

jmohdyusof:
What was the code change? I looked at the section of the code you said was causing problems but was having trouble seeing any issues with formatting.

Jamaludin Mohd Yusof, PhD
Future Architectures and Applications
CCS-7 Applied Computer Science Group
Phone: (505) 665-2870
Fax  : (505) 665-4939
Email: jamal@lanl.gov<mailto:jamal@lanl.gov>

On Apr 18, 2019, at 3:22 PM, Jean-Luc Fattebert <notifications@github.com<mailto:notifications@github.com>> wrote:


I changed the code to avoid the issue. Now there seems to be another issue I don't understand.

jeanlucf22:
I used a temporary variable with a shorter name. There was a function call with an argument "A->distribution_mode" which indent would split over two lines, and then merge again at the next application of indent. It was at teh beginning of the very first function in bml_transpose_new_ellblock_typed.c. I now store that argument into a temporary variable before passing it to the function call.

nicolasbock:
Thanks @jeanlucf22 for fixing the indentation! By the way, I see the same bug with indent you mention. Strange, I had never run into this before.

I restarted CI to test your latest version.

nicolasbock:
Rebased

nicolasbock:
Thanks!

PR #267, opened 2019-06-27
Fix `const` types in `bml_add_norm`

nicolasbock:
This change adjusts the types in `bml_add_norm` so that the pointer
`A` is `const`, but the value of `A` is mutable, and both the pointer
and the value of `B` is `const`.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/267)

### DIFF
--- a/src/C-interface/bml_add.c
+++ b/src/C-interface/bml_add.c
@@ -63,11 +63,11 @@ bml_add(
  */
 double
 bml_add_norm(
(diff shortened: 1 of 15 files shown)


jeanlucf22:
I see things like: "bml_matrix_dense_t const *const B"
Shouldn't it be: "const bml_matrix_dense_t* const B" instead?

PR #270, opened 2019-06-27
Make `emacs` binary configurable

nicolasbock:
For local testing it can be convenient to be able to set the `emacs`
binary explicitly. This change introduces the environment variable
`EMACS` to specify which `emacs` binary to use.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/270)


nicolasbock:
Thanks!

PR #271, opened 2019-06-27
Fix incorrect indentation

nicolasbock:
Some of the C header files were using incorrect indentation.

Signed-off-by: Nicolas Bock <nicolasbock@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/271)

PR #273, opened 2019-06-27
Const 4

nicolasbock:
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/273)

Depends-On: #271

### DIFF
--- a/src/C-interface/bml_add.h
+++ b/src/C-interface/bml_add.h
@@ -25,9 +25,9 @@ void bml_add_identity(
     const double threshold);
 
 void bml_scale_add_identity(
-    bml_matrix_t * A,
-    const double alpha,
-    const double beta,
-    const double threshold);
+    bml_matrix_t * const A,
+    double const alpha,
+    double const beta,
+    double const threshold);
 
 #endif
--- a/src/C-interface/bml_add.h
+++ b/src/C-interface/bml_add.h
@@ -26,8 +26,8 @@ void bml_add_identity(
 
 void bml_scale_add_identity(
     bml_matrix_t * const A,
-    double const alpha,
-    double const beta,
-    double const threshold);
+    const double alpha,
+    const double beta,
+    const double threshold);
 
 #endif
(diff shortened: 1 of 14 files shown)


src/C-interface/bml_add.h
```diff
@@ -14,7 +14,7 @@ void bml_add(
 
 double bml_add_norm(
     bml_matrix_t * const A,
-    bml_matrix_t const * const B,
```
jeanlucf22: Shouldn't the first cont be in front of bml_matrix_t?

nicolasbock: According to section 6.7 of the C17 draft the order is not relevant and could be `const bml_matrix_t` or `bml_matrix_t const`. I can switch it around to read `const bml_matrix_t`. I agree, that this order somehow reads more naturally.

jmohdyusof: It would also be more readable if the * is attached (without a space) to one or other of the variables declarations.

nicolasbock: Yes, tru @jmohdyusof . I haven't figured out yet how to tell `indent` to do that. I'll keep trying.

jmohdyusof: Hi Nick, it looks like you got the const *const to work, but the other declarations still have 'floating' asterisks? Not sure how much you should worry about this.

nicolasbock: Yes, I played around with the command line arguments of `indent` but so far I haven't found a combination that places the `*` correctly in all cases. We can ignore it maybe for now and keep playing around with the arguments and then fix it if we find the right combination.

jmohdyusof: In my OMP offload versions I removed *all* the const declarations from the array pointers which were being written to (due to compiler errors), but this is probably overkill. I will see if the declarations you have allow compilation with the offload.

nicolasbock:
I need a review to merge this though @jmohdyusof and @jeanlucf22

src/C-interface/bml_add.h
nicolasbock: Perfect! Let me know how that goes! If this PR is doing the trick then we can go ahead and change the other uses of `const`.

nicolasbock:
Thanks!

PR #276, opened 2019-07-02
Fix `const` types in `bml_allocate`

nicolasbock:
This change adjusts the types in `bml_allocate` so that the const type
modifier applies to the correct part of the type.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/276)

### DIFF
--- a/src/C-interface/bml_allocate.c
+++ b/src/C-interface/bml_allocate.c
@@ -21,12 +21,12 @@
  *
  * \ingroup allocate_group_C
  *
- * \param A Matrix
- * \return >0 if allocated, else -1
+ * \param A[in,out] Matrix
+ * \return \f$ > 0 \f$ if allocated, else -1
  */
 int
 bml_allocated(
-    const bml_matrix_t * A)
+    bml_matrix_t * A)
 {
     return bml_get_N(A);
 }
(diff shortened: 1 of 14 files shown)


jeanlucf22:
Why would we use 'const' in bml_deallocate()? This function IS changing the pointer...

nicolasbock:
@jeanlucf22 , @jmohdyusof , please have a look. I removed the `const` qualifiers in the allocate functions.

src/C-interface/bml_allocate.c
```diff
@@ -21,12 +21,12 @@
  *
  * \ingroup allocate_group_C
  *
- * \param A Matrix
- * \return >0 if allocated, else -1
+ * \param A[in,out] Matrix
+ * \return \f$ > 0 \f$ if allocated, else -1
```
jeanlucf22: What is \f$ supposed to mean here?

nicolasbock: That's doxygen's way to include a LaTeX expressions. I figured it would look nicer to typeset that properly :smile:

nicolasbock: http://www.doxygen.nl/manual/formulas.html

nicolasbock:
Thanks!

PR #308, opened 2019-07-10
Add test for bml_print_bml_matrix functions

jeanlucf22:
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/308)

### DIFF
--- a/src/C-interface/bml_allocate.c
+++ b/src/C-interface/bml_allocate.c
@@ -370,7 +370,6 @@ bml_random_matrix(
     int M,
     bml_distribution_mode_t distrib_mode)
 {
-    LOG_DEBUG("random matrix of size %d\n", N);
     switch (matrix_type)
     {
         case dense:
--- a/src/C-interface/bml_export.c
+++ b/src/C-interface/bml_export.c
@@ -35,7 +35,6 @@ bml_export_to_dense(
     bml_matrix_t * A,
     bml_dense_order_t order)
 {
-    LOG_DEBUG("Exporting bml matrix to dense\n");
     switch (bml_get_type(A))
     {
         case dense:
--- a/src/C-interface/bml_utilities.c
+++ b/src/C-interface/bml_utilities.c
@@ -365,7 +365,6 @@ bml_print_dense_vector(
     int i_l,
     int i_u)
 {
-    LOG_DEBUG("printing vector [%d:%d]\n", i_l, i_u);
     switch (matrix_precision)
     {
         case single_real:
--- a/tests/C-tests/CMakeLists.txt
+++ b/tests/C-tests/CMakeLists.txt
@@ -17,6 +17,7 @@ set(SOURCES_TYPED
   multiply_matrix_x2_typed.c
   normalize_matrix_typed.c
   norm_matrix_typed.c
+  print_matrix_typed.c
   scale_matrix_typed.c
   set_row_typed.c
   submatrix_matrix_typed.c
@@ -58,6 +59,7 @@ add_executable(bml-test
   multiply_matrix_x2.c
   normalize_matrix.c
   norm_matrix.c
+  print_matrix.c
   scale_matrix.c
   set_row.c
   submatrix_matrix.c
@@ -102,6 +104,7 @@ foreach(N
     multiply_x2
     norm
     normalize
+    print
     scale
     set_row
     submatrix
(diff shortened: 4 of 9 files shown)


jeanlucf22:
Working on it. It passes all the tests on my local machine, so not an obvious bug...

jmohdyusof:
I was trying to find where the print_matrix functions are defined. Which branch/file are they in?

jeanlucf22:
Look at bml_utilities_* files.

jmohdyusof:
The complex types are being cast to float/double respectively? Is that correct? Otherwise it is hard to see how it would differ since both use export_to_dense and print_dense.

jeanlucf22:
Are you referring to the test function?

jmohdyusof:
In bml_utilities, the complex matrices are initialized as real.

jeanlucf22:
With this test, I am trying to check what is printed out to stdout by redirecting it to a file, which is printed out with a small number of digits. So precision is not really the issue. The main issue is to read complex numbers that look like 0.840+0.000i and compare those with _Complex type values. I am using a pretty loose tolerance of 10^-3 because of the small number of digits written out by function to be tested.

nicolasbock:
In my local tests your tests are passing. Strange.

nicolasbock:
I just realized that the `bml_random_matrix()` function only produces real valued matrices. Maybe not very useful for complex types.

jeanlucf22:
One thing I noticed is that in the CI testing, two tests can be run simultaneously, and conflicts could happen with two tests trying to access the same file. My last fix is taking care of that.

jeanlucf22:
> I just realized that the `bml_random_matrix()` function only produces real valued matrices. Maybe not very useful for complex types.

Indeed. I believe some tests need to be revisited and strengthened.

jmohdyusof:
My point was that you are passing a complex matrix to a real type, so you are inserting extra zeros into the real matrix (the imaginary parts will appear in the alternating real outputs)

nicolasbock:
> One thing I noticed is that in the CI testing, two tests can be run simultaneously, and conflicts could happen with two tests trying to access the same file. My last fix is taking care of that.

Ah, good catch. We are running the tests in parallel (see https://github.com/lanl/bml/blob/master/build.sh#L197). So you need to make sure you don't introduce race conditions. But it looks like your approach of naming the temporary files works. You could alternatively use `mktemp()`.

jeanlucf22:
Still failing...

nicolasbock:
The interesting thing is that the `dense` type passes.

nicolasbock:
@jeanlucf22 I think I found the problem. Do you mind if I push my version?

jeanlucf22:
Go ahead!

nicolasbock:
Looks like an improvement. At this point the valgrind tests are failing, but the other ones aren't. I'll check what the issue with valgrind is.

nicolasbock:
I don't quite understand the valgrind errors yet. They seem to be in `printf` for example, or they might be triggered inside `printf` by the arguments we are passing. It looks like your new test is uncovering some issues we hadn't noticed so far @jeanlucf22 ! Thanks! Once we fix those the code will be better :smile:

jeanlucf22:
There was an obvious memory leak that I just fixed and pushed. Let see if there is more...

nicolasbock:
Thanks @jeanlucf22 ! I think I found the rest. I just pushed another attempt.

nicolasbock:
@jeanlucf22 I haven't fixed the last test and simply added some logic to skip it, but now all tests pass :smile:

In a follow up, we need to address the situation where the compiler does not support complex types and we map the complex types to the corresponding real types so we don't have to change the API. However, your test shows that this is not working everywhere correctly.

I don't want to block this PR until we fix `BML_COMPLEX=no` and I think it's fine to add the conditional skip for now.

jeanlucf22:
Thanks @nicolasbock ! What was the main issue making the test failing for ellpack single_real for instance? Was the  fflush(stdout) that you added the key? Restoring stdout certainly makes it cleaner, but was it also required to make it work?

nicolasbock:
No problem :smile: The issue was that when we build with `CMAKE_BUILD_TYPE=Debug` the `export` code printed a line to `stdout` which made the parsing fail. I removed the lines since they don't really add anything of value anyhow. I suppose we could have left them also but print to stderr instead.

You are right that restoring `stdout` wasn't really an issue.

PR #330, opened 2020-01-17
Ellblock blas

jeanlucf22:
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/330)


cmake/FindXSMM.cmake
```diff
@@ -0,0 +1,36 @@
+# - Find the XSMM library
```
nicolasbock: Are we using this file anywhere?

jeanlucf22: CMakeLists.txt, line 359

nicolasbock: Oh how cool. I didn't realize that you could extend CMake that way. Thanks!

PR #347, opened 2020-02-11
Fix memory leak

nicolasbock:
We should investigate whether the call to `bml_multiply_x2_ellsort` is
necessary or could be written slightly differently so we don't have to
free the trace which is not used here.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/347)

### DIFF
--- a/src/C-interface/ellsort/bml_multiply_ellsort_typed.c
+++ b/src/C-interface/ellsort/bml_multiply_ellsort_typed.c
@@ -46,6 +46,8 @@ void TYPED_FUNC(
     double ONE = 1.0;
     double ZERO = 0.0;
 
+    void *trace = NULL;
+
     if (A == NULL || B == NULL)
     {
         LOG_ERROR("Either matrix A or B are NULL\n");
@@ -53,7 +55,7 @@ void TYPED_FUNC(
 
     if (A == B && alpha == ONE && beta == ZERO)
     {
-        TYPED_FUNC(bml_multiply_x2_ellsort) (A, C, threshold);
+        trace = TYPED_FUNC(bml_multiply_x2_ellsort) (A, C, threshold);
     }
     else
     {
@@ -64,7 +66,7 @@ void TYPED_FUNC(
 
         if (A != NULL && A == B)
         {
-            TYPED_FUNC(bml_multiply_x2_ellsort) (A, A2, threshold);
+            trace = TYPED_FUNC(bml_multiply_x2_ellsort) (A, A2, threshold);
         }
         else
         {
@@ -82,6 +84,8 @@ void TYPED_FUNC(
 
         bml_deallocate_ellsort(A2);
     }
+
+    bml_free_memory(trace);
 }
 
 /** Matrix multiply.


jmohdyusof:
I assume trace isn't needed since it is only used because of the way bml_multiply_x2 is defined. If we follow the suggestion on the earlier thread and make multiply_x2 have the same signature as multiply_AB, with no trace returned then the problem goes away. We define trace_mult_x2 to have the form with the trace returned.

src/Fortran-interface/bml_getters_m.F90
```diff
@@ -117,6 +117,7 @@ subroutine bml_get_row_single_real(a, i, row)
     ptr = bml_get_row_C(a%ptr, i - 1)
     call c_f_pointer(ptr, row_ptr, [bml_get_N(a)])
     row = row_ptr
+    deallocate(row_ptr)
```
nicolasbock: This fixes the valgrind test. This suggests that the assignment in https://github.com/lanl/bml/pull/347/commits/39050c9f2e4c31b0af8a6aa595ef8c6401918115#diff-fc698c8987a3d4f990baad818a46a573R119 is copying the data instead of assigning a reference to it. @cnegre @jeanlucf22 @jmohdyusof @suemni what do you think?

nicolasbock:
It looks like @suemni introduced the `trace` in f87af50652b079c2386b374d920a17d25b6440dc but I presume she had a good reason to do so. @suemni could you comment? Should we keep that additional `bml_free_memory` calls or remove the `trace` allocation in the `x2` function?

nicolasbock:
I split the second commit off into a separate PR: #348

junghans:
# [Codecov](https://codecov.io/gh/lanl/bml/pull/347?src=pr&el=h1) Report
> Merging [#347](https://codecov.io/gh/lanl/bml/pull/347?src=pr&el=desc) into [master](https://codecov.io/gh/lanl/bml/commit/ae789f6ed8163a2d2e2639541b311b896a2ca82a&el=desc) will **increase** coverage by `0%`.
> The diff coverage is `n/a`.

[Impacted file tree graph](https://codecov.io/gh/lanl/bml/pull/347?src=pr&el=tree)

```diff
@@          Coverage Diff           @@
##           master    #347   +/-   ##
======================================
  Coverage      62%     62%           
======================================
  Files         300     300           
  Lines       12855   12857    +2     
  Branches     1067    1067           
======================================
+ Hits         8021    8023    +2     
  Misses       4574    4574           
  Partials      260     260           
```


| [Impacted Files](https://codecov.io/gh/lanl/bml/pull/347?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [src/C-interface/ellblock/bml\_copy\_ellblock\_typed.c](https://codecov.io/gh/lanl/bml/pull/347/diff?src=pr&el=tree#diff-c3JjL0MtaW50ZXJmYWNlL2VsbGJsb2NrL2JtbF9jb3B5X2VsbGJsb2NrX3R5cGVkLmM=) | `56% <0%> (ø)` | :arrow_up: |

------

[Continue to review full report at Codecov](https://codecov.io/gh/lanl/bml/pull/347?src=pr&el=continue).
> **Legend** - [Click here to learn more](https://docs.codecov.io/docs/codecov-delta)
> `Δ = absolute <relative> (impact)`, `ø = not affected`, `? = missing data`
> Powered by [Codecov](https://codecov.io/gh/lanl/bml/pull/347?src=pr&el=footer). Last update [ae789f6...d18fb33](https://codecov.io/gh/lanl/bml/pull/347?src=pr&el=lastupdated). Read the [comment docs](https://docs.codecov.io/docs/pull-request-comments).

suemni:
The algorithm for X**2 involved calculating the trace of X and X**2. It may be used also by some of the PROGRESS solvers. But deallocating of the trace array needs to be done.


- Sue

jeanlucf22:
The trace is needed. See:
https://github.com/lanl/qmd-progress/blob/master/src/prg_sp2_mod.F90
line 85 for example

jmohdyusof:
i guess it would make sense semantically to use a command called trace_mult_x2, or to have the the multiply_AB also return a trace..

nicolasbock:
Thanks! I'll merge this change so the memory leak is fixed. We can discuss later if and how to change the API as @jmohdyusof suggests.

PR #348, opened 2020-02-11
Fix memory leak for getter

nicolasbock:

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/348)

### DIFF
--- a/src/Fortran-interface/bml_getters_m.F90
+++ b/src/Fortran-interface/bml_getters_m.F90
@@ -41,6 +41,7 @@ contains
     ptr = bml_get_diagonal_C(a%ptr)
     call c_f_pointer(ptr, diagonal_ptr, [bml_get_N(a)])
     diagonal = diagonal_ptr
+    deallocate(diagonal_ptr)
 
   end subroutine bml_get_diagonal_single_real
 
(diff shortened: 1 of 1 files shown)


src/Fortran-interface/bml_getters_m.F90
```diff
@@ -117,6 +117,7 @@ subroutine bml_get_row_single_real(a, i, row)
     ptr = bml_get_row_C(a%ptr, i - 1)
     call c_f_pointer(ptr, row_ptr, [bml_get_N(a)])
     row = row_ptr
+    deallocate(row_ptr)
```
nicolasbock: This fixes the valgrind test. This suggests that the assignment in https://github.com/lanl/bml/pull/348/files#diff-fc698c8987a3d4f990baad818a46a573R119 is copying the data instead of assigning a reference to it. @cnegre @jeanlucf22 @jmohdyusof @suemni what do you think?

cnegre:
Sounds good to me!

src/Fortran-interface/bml_getters_m.F90
nicolasbock: If it's really a data copy operation, then it's less efficient than I had hoped. Does anyone know how to get the Fortran compiler to use the reference when converting a `real(C_FLOAT), pointer` to a `real(C_FLOAT), allocatable (:)`?

nicolasbock:
Reviews please :smile: @cnegre @suemni @jmohdyusof @tokshgithub

PR #372, opened 2020-04-22
cmake: export bml target

junghans:
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/372)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 3.10)
 
 message(STATUS "CMake version ${CMAKE_VERSION}")
 
@@ -220,7 +220,7 @@ if(BML_OPENCL)
   include(FindOpenCL)
 endif()
 
-SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+SET(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 
 set(BML_MAGMA FALSE CACHE BOOL "Whether to use MAGMA library")
 if(BML_MAGMA)
@@ -384,7 +384,7 @@ if(NOT (LAPACK_FOUND OR NOBLAS))
   endif()
 endif()
 
-include(${CMAKE_SOURCE_DIR}/cmake/bmlCheckCFortranFunctionExists.cmake)
+include(${PROJECT_SOURCE_DIR}/cmake/bmlCheckCFortranFunctionExists.cmake)
 if(MAGMA_FOUND)
   set(CMAKE_REQUIRED_LIBRARIES
     ${MAGMA_LIBRARIES})
@@ -491,7 +491,7 @@ set(BML_COMPLEX TRUE
   CACHE BOOL "Whether to build complex types")
 
 if(BML_COMPLEX)
-  try_compile(SUPPORTS_COMPLEX ${CMAKE_BINARY_DIR} ${CMAKE_SOURCE_DIR}/cmake/test_complex.c
+  try_compile(SUPPORTS_COMPLEX ${CMAKE_BINARY_DIR} ${PROJECT_SOURCE_DIR}/cmake/test_complex.c
     OUTPUT_VARIABLE COMPLEX_OUTPUT
     )
   message(STATUS "Compile output: ${COMPLEX_OUTPUT}")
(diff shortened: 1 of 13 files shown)


nicolasbock:
## Danger has errored
[!] Invalid `Dangerfile` file: 404 Not Found
```
 #  from Dangerfile:18
 #  -------------------------------------------
 #  # reviewers.
 >  mention.run(2, [], [])
 #  
 #  -------------------------------------------
```

  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

junghans:
@nicolasbock, I ran
```
CC=clang-9 CXX=clang++-9 FC=gfortran BUILD_SHARED_LIBS=yes BML_OPENMP=no BML_INTERNAL_BLAS=no BML_VALGRIND=no OMP_NUM_THREADS=4 CMAKE_BUILD_TYPE=Debug VERBOSE_MAKEFILE=yes PARALLEL_TEST_JOBS=2 ./build.sh testing
```
locally, but cannot trigger that CircleCI issue.

junghans:
@nicolasbock the remaining failure is due to the fact that docs image is a bit outdated, could you please update?

nicolasbock:
> @nicolasbock the remaining failure is due to the fact that docs image is a bit outdated, could you please update?

Sure, I'll have a look. Thanks for the fixes!

CMakeLists.txt
```diff
@@ -220,7 +220,7 @@ if(BML_OPENCL)
   include(FindOpenCL)
 endif()
 
-SET(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+SET(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
```
nicolasbock, line 223: Why change those to `PROJECT_SOURCE_DIR`? That's semantically different than using `CMAKE_SOURCE_DIR`. In our case it ends up being the same, but if we were ever to move the `project()` call to another `CMakeLists.txt` this would break.

junghans: We are trying to do a `LATTE` super-build with:
```cmake
cmake_minimum_required(VERSION 3.9)
project(latte_super)

add_subdirectory(bml)
add_subdirectory(qmd-progress)
option(PROGRESS "Use progress library" ON)
add_subdirectory(latte/cmake)
```
and if you include `bml` like this `CMAKE_SOURCE_DIR` points to wrong directory, while `PROJECT_SOURCE_DIR` is correct.

junghans: The only place where I wasn't sure is in `test-getter`, but there we could link against `BML::bml` as well.

jmohdyusof: Just a comment that in general the CMake seems to use keywords that do not correspond to the system variables which forces stuff like
export CUDA_TOOLKIT_ROOT_DIR=${CUDADIR}
Is there a reason not to use the env variables directly?

junghans: bml's cmake build system needs a deep scrub, since cmake-3.8 cuda is a first class citizen, I haven't really touched that part in this PR, but the `CUDA_TOOLKIT_ROOT_DIR` environment variable has always been CMake's variable to detect cuda.

jmohdyusof: OK, I generally find the CMake build system hard to follow at the best of times. It would help a lot if our example build scripts were even close to a working example.

nicolasbock: The bml is a dependency of LATTE. I would expect that LATTE adjusts to the bml and not the other way around. I am not saying that I see any serious reason to move the `project()` call, but it does feel wrong to make that change here.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.7)
+cmake_minimum_required(VERSION 3.10)
```
nicolasbock, line 1: Is this version available on all systems we are building the library on?

junghans: We are following LAMMPS lead here, see https://github.com/lammps/lammps/pull/1723#issuecomment-542280424 for why CMake-3.10 is reasonable beside the fact that it was released in Nov. 2017.

nicolasbock: `trusty` (Ubuntu 14.04) is EOL April 2022 for example and ships with cmake-2.8.12.2.

nicolasbock: Ok, just making sure. I don't know what's running on current supercomputers these days :smile:

CMakeLists.txt
junghans: @nicolasbock well this applies to any time, we want to use `add_subdirectory(bml)` for any reason, latte is just one example. It seems to me most CMake projects have converged to allow that the inclusion for their projects as subdirectories these days, so I followed that.

CMakeLists.txt
junghans: If you want to support `trusty` go for it, but 6 years is a long time ago. I would say if somebody runs `trusty` they need to stick to `bml-1.3.1` instead of the latest master.

nicolasbock: As I said, I don't know what OSes currently used systems at the labs are running. The last time I checked they were at best "conservative" in their upgrade schedule.

The main point I was trying to make is that we need to agree that we will drop support for CMake < 3.10.

CMakeLists.txt
nicolasbock: Could you then at least add a comment so that someone does not refactor the `project()` call into a subdirectory?

CMakeLists.txt
junghans: @suemni @cnegre your call.

CMakeLists.txt
junghans: Sorry, but why would somebody try to move the `project()` line?

nicolasbock: Because you can. The comment is just insurance so that in case a year down the road someone goes wild with refactoring the CMake scripts we remember that we should not move that call.

junghans: Well, if I move a random line of CMake code into a different file, it will most likely break something, and the `project()` line is no exception.

nicolasbock: > PROJECT_SOURCE_DIR
Top level source directory for the current project.
This is the source directory of the most recent project() command.

https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html

However, that's incorrect because scoping in CMake works only down, not up (which I couldn't find good documentation for) and a `project()` call in `add_subdirectory()` does not propagate back to the calling scope.

:+1: on the change.

PR #376, opened 2020-04-30
Update bml_allocate.c

tokshgithub:
modified HAVE_POSIX_MEMALIGN macro for first touch

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/376)

### DIFF
--- a/src/C-interface/bml_allocate.c
+++ b/src/C-interface/bml_allocate.c
@@ -57,6 +57,7 @@ bml_allocate_memory(
 #elif defined(HAVE_POSIX_MEMALIGN)
     char *ptr;
     posix_memalign((void **) &ptr, MALLOC_ALIGNMENT, size);
+#pragma omp simd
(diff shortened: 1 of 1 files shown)


src/C-interface/bml_allocate.c
```diff
@@ -54,6 +54,7 @@ bml_allocate_memory(
 #elif defined(HAVE_POSIX_MEMALIGN)
     char *ptr;
     posix_memalign((void **) &ptr, MALLOC_ALIGNMENT, size);
+    #pragma omp parallel for simd
```
nicolasbock, line 57: I think the change is fine, however, this raises the compiler requirement to `OpenMP >= 4.0`, which means `gcc >= 4.9`. Right now we are still testing with `gcc-4.7`. In terms of Ubuntu this excludes only Trusty (14.04) which is probably not a big issue.

nicolasbock (approved):
👍🏾

PR #392, opened 2020-05-07
Do not use pre-built images for CI

nicolasbock:
This will allow us to modify the build environment from git since we
don't depend on our own images anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/392)


nicolasbock:
Note, the only reason we might want to use a pre-built image is turnaround time for CI.

PR #398, opened 2020-06-15
Add memory pool for ellblock

jeanlucf22:
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/398)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -519,6 +519,11 @@ if(NOT (NOBLAS))
   endif()
 endif()
 
+if(BML_ELLBLOCK_MEMPOOL)
+  message(STATUS "Use memory pool for elllock")
+  add_definitions(-DBML_ELLBLOCK_USE_MEMPOOL)
+endif()
+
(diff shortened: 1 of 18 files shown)


jeanlucf22:
@tokshgithub I am trying to allocate memory upfront to avoid a memory allocation for every single block when they come into play. Open to suggestions if you see flaws with the way I do it.

nicolasbock:
@jeanlucf22 I'd be interested in any performance comparisons you might have. How much faster is pre-allocating?

jeanlucf22:
@nicolasbock I will check performance now that I fixed all the bugs. If performance is much better, I will remove the [WIP] label and propose to merge this branch. Otherwise, I may put in compile time flags to choose between old way and new way of allocating memory

nicolasbock:
Perfect, thanks!

jeanlucf22:
It looks like using a memory pool doesn't affect performance much. But this is still useful, potentially for GPU offload.
Thus I made it a compile time option to use or not use the memory pool. It is ON by default.

jeanlucf22:
Ready to be merged!

PR #401, opened 2020-07-20
Ellblock optimization

jeanlucf22:
Optimize small matrix multiplications when libxsmm is not available

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/401)

### DIFF
--- a/src/C-interface/ellblock/bml_multiply_ellblock_typed.c
+++ b/src/C-interface/ellblock/bml_multiply_ellblock_typed.c
@@ -25,6 +25,148 @@
 #include <omp.h>
 #endif
 
+void TYPED_FUNC(
+    bml_multiply_block1) (
+    REAL_T * mat0,
+    REAL_T * mat1,
+    const int ld,
+    REAL_T * x,
+    const int bsizeib,
+    const int bsizekb)
+{
+    REAL_T *xk = x;
+    for (int ii = 0; ii < bsizeib; ii++)
+    {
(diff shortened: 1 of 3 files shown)


jeanlucf22:
@nicolasbock The option of using libxsmm has been in for a while, see build.sh

nicolasbock:
Right @jeanlucf22 . What I meant to say is that the lines pertaining to `libxsmm` don't seem to be connected with the subject of the commit, i.e. "Unroll inner loop in functions". I presume that the `libxsmm_init()` and `libxsmm_finalize()` calls should be in the code but could you elaborate a little in the commit message to explain why those lines are added now and were missing before?

If the code didn't work before we are missing some test coverage in our CI. If that's the case, we should add a test for libxsmm. I can certainly help you there.

Thanks!

jeanlucf22:
@nicolasbock true, the libxsmm_init and finalize don't really belong in this PR. I'll split the PR. It is not incorrect when they are missing. The init is just "an explicit initialization routine helps to avoid lazy initialization overhead when calling LIBXSMM for the first time".

jeanlucf22:
@nicolasbock as far as testing is concerned, the use of libxsmm is not tested. That would require an extra build with libxsmm, or more...

jeanlucf22:
@nicolasbock libxsmm related stuff has been removed (and merged with PR #402). Now ready to merge if you are happy with the changes

nicolasbock:
That looks great @jeanlucf22 ! Thanks!

PR #402, opened 2020-07-21
Fix a couple of minor issues with libxsmm

jeanlucf22:
* call libxsmm_init() and libxsmm_finalize()

* add link to blas after libxsmm in cmake

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/402)


nicolasbock:
Thanks @jeanlucf22 !

PR #415, opened 2020-12-19
Implement initial distributed functionalities

jeanlucf22:
Implement infrastructure and first functionalities for new format
"distributed2d": allocation. import. export.
If build with MPI, these functionalities will be tested for all
"serial" formats as building blocks.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/415)


nicolasbock:
## Danger has errored
[!] Invalid `Dangerfile` file: 429 Too Many Requests. Updating the Danger gem might fix the issue. Your Danger version: 7.0.1, latest Danger version: 8.2.1

```
 #  from Dangerfile:18
 #  -------------------------------------------
 #  # reviewers.
 >  mention.run(2, [], [])
 #  
 #  -------------------------------------------
```

  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

jeanlucf22:
@nicolasbock could you help me with the CI? I would like to add tests with "BML_MPI=yes"

suemni (approved):

nicolasbock:
@jeanlucf22 Of course. I am looking at how to best add this to CI.

nicolasbock:
Hi @jeanlucf22 ! I merged #421 which adds MPI libraries to the CI container image. You can build the code in the container now with
```
# BML_MPI=yes ./build.sh compile
```
I haven't done any further testing. But it appears to compile 😄 Let me know if that works for you.

Best,

Nick

jeanlucf22:
Hi @nicolasbock ! Looks like I am now able to add an MPI test to the CI, but it fails to build (cannot find MPI library)

nicolasbock:
Thanks for the added tests. I'll have a closer look.

nicolasbock:
## Danger has errored
[!] Invalid `Dangerfile` file: 429 too many requests. Updating the Danger gem might fix the issue. Your Danger version: 7.0.1, latest Danger version: 8.2.1

```
 #  from Dangerfile:18
 #  -------------------------------------------
 #  # reviewers.
 >  mention.run(2, [], [])
 #  
 #  -------------------------------------------
```

  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

jmohdyusof:
Why do we have a separate function XXX_distributed2d when the existing methods take the distrib_mode argument? Why not implement distrib_mode=2d within those existing functions?

jeanlucf22:
> Why do we have a separate function XXX_distributed2d when the existing methods take the distrib_mode argument? Why not implement distrib_mode=2d within those existing functions?

(1) We could have other "distributed" types. Being distributed is not a type in itself. "distributed2d" is a new type  (2) I implemented it the way I think is best, but I am open to suggestions (3) I did not feel like I had to use an argument that had been put there based on an unfinished design. I actually would like to take out the distrib_mode argument of the "serial" functions at some point

jeanlucf22:
> > Why do we have a separate function XXX_distributed2d when the existing methods take the distrib_mode argument? Why not implement distrib_mode=2d within those existing functions?
> 
> (1) We could have other "distributed" types. Being distributed is not a type in itself. "distributed2d" is a new type (2) I implemented it the way I think is best, but I am open to suggestions (3) I did not feel like I had to use an argument that had been put there based on an unfinished design. I actually would like to take out the distrib_mode argument of the "serial" functions at some point

In addition, the way I am doing requires no changes to any of the existing "serial" types while allowing to use all of these existing types

jmohdyusof:
You are using the distrib_mode value as the switch to call the _distributed2d version, why not simply put the switch inside the existing function (and you can still separate the code as needed). Most of the existing code I touched (and sometimes fixed) is already set up to at least in principle work with distributed1d since it has start and end row. It would make more sense to me to modify the existing functions, using eg. 'serial' '1d' and '2d' as values for distrib_mode.

jeanlucf22:
> You are using the distrib_mode value as the switch to call the _distributed2d version, why not simply put the switch inside the existing function (and you can still separate the code as needed). Most of the existing code I touched (and sometimes fixed) is already set up to at least in principle work with distributed1d since it has start and end row. It would make more sense to me to modify the existing functions, using eg. 'serial' '1d' and '2d' as values for distrib_mode.

I would need to touch every single function of all the serial formats. In addition, the "start" and "end" rows have not been used consistently and are not tested at all. I would suggest to drop them all.

jmohdyusof:
As I said, I fixed all the functions I touched to use start and end rows. If you want to remove them are you proposing that 1d distribution will never be used? I guess this is a discussion for more people than just the 2 of us?

jeanlucf22:
Indeed I would love to hear feedback from others. With this PR, I was actually hoping for feedback!

As far as a 1D distributed option, the code that is in was introduced before my time and I don't know anybody who is pursuing it. Note that we could keep it: my "distributed2d" design does not exclude it.

jmohdyusof:
I guess I will just register my preference for having a consolidated 'switch' on distrib_mode with multiple options, rather than having the 'if' 'then' 'else' 'switch' that you have now. Also, is 2d only intended for dense format or is it for all? At the function level, the parallel struct just has to pass the row numbers to make the 1d 'work' (it isn't clear to me where the communication happens). I can envision a moderate size range for N, M where 1d distribution makes sense because 2d would generate too little local work, but perhaps that is incorrect intuition?

jeanlucf22:
You can use all "serial" formats with my "distributed2d". This is actually an important point in my design. 
If you build with MPI and run the tests, you will see how it will run all the tests implemented so far for all the "serial" formats.

Re: "it isn't clear to me where the communication happens"
I believe it was never implemented... Again, the 1d distribution was never completed. I don't know if it was fully thought through.

jmohdyusof:
Yes, I understand that you can use 'serial' but the control flow is complicated: why does distrib_mode get passed to _distributed2d if it already made the switch? Is bml_zero_matrix_distributed2d implementing only dense or all the matrix_types 

#ifdef DO_MPI
    if (distrib_mode == distributed)
        return bml_zero_matrix_distributed2d(matrix_type, matrix_precision, N,
                                             M, distrib_mode);
    else
#endif
    {
        switch (matrix_type)
        {
            case dense:
                return bml_zero_matrix_dense(matrix_precision,
                                             matrix_dimension, distrib_mode);
                break;

jeanlucf22:
@jmohdyusof This was an unused argument. I just removed it. Thanks for catching that.

jmohdyusof:
OK, I guess we will wait for other comments. So long as we have a coherent branching structure/control flow I am not too worried about the actual partitioning of the code. Does this make sense?

jeanlucf22:
@jmohdyusof Makes sense. Thanks for your feedback.

nicolasbock:
Rebased

jeanlucf22:
Converted to draft until we have time to discuss it as a team

jeanlucf22:
@nicolasbock The MPI test still cannot find an MPI library. Should it be added to ci-images/docs/Dockerfile?

nicolasbock:
@jeanlucf22 The dockerfile does not directly affect the CI testing. There is an intermediate step necessary and we need to push the new image to the DockerHub registry for it to take effect. I am looking at the CI failure right now to see what the problem is. Thanks!

nicolasbock:
@jeanlucf22 I think the issue was missing `-I ...` and `-l` directives. The new commit I added fixes the issue locally for me. Let's hope GitHub is as convinced 😉

nicolasbock:
@jeanlucf22 This is looking better I think. For some reason we weren't using the current CI image in our tests. I re-ran the container build step on `master` and now the MPI tests are passing.

nicolasbock (approved):
Thanks @jeanlucf22 !

PR #435, opened 2020-12-26
Automate API doc building

nicolasbock:


suemni (approved):

jmohdyusof:
Don't we need sphinx_rtd_theme instead of alabaster for the html theme?

nicolasbock:
> Don't we need sphinx_rtd_theme instead of alabaster for the html theme?

Yes possibly. I am still trying to get the PR to build on readthedocs to see what the result looks like.

jmohdyusof:
The top-level readme is not included unless you add it to the source/index.rst. Note also that you cannot specify relative paths (above the source directory) easily in sphinx, so you need to use an include file, or move the doc to within source (In other projects I have a source/usage subdir for this).

nicolasbock:
Locally the docs are building for me. I submitted lanl/bml#574 to fix some stylistic issues and hopefully that will address the build failure on rtd (https://readthedocs.org/projects/basic-matrix-library/builds/15891598/)

PR #439, opened 2021-02-05
Add build script for Summit (OLCF)

jeanlucf22:
We can save some time to users and developers by adding working build
scripts for a few platforms

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/439)


nicolasbock (approved):
Thanks!

PR #440, opened 2021-02-11
Add precommand and precommand argument in test runs

jeanlucf22:
This is necessary for example on Summit at OLCF to run on a compute
node using jsrun. This is now also used for MPI tests, replacing hard
coded mpirun

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/440)


suemni (approved):

build.sh
```diff
@@ -48,10 +48,15 @@ EOF
     echo "FC                     Path to Fortran compiler    (default is ${FC})"
     echo "BML_OPENMP             {yes,no}                    (default is ${BML_OPENMP})"
     echo "BML_MPI                {yes,no}                    (default is ${BML_MPI})"
+    echo "BML_MPIEXEC_EXECUTABLE Command to prepend MPI tests (default is ${BML_MPIEXEC_EXECUTABLE})"
+    echo "BML_MPIEXEC_NUMPROCS_FLAG Flags to specify number of MPI tasks (default is ${BML_MPIEXEC_NUMPROCS_FLAG})"
```
nicolasbock, line 52: Would it make sense to fold this option into the `BML_MPIEXEC_PREFLAGS` option as well?

jeanlucf22: I am following cmake here: these are the variables cmake defines and uses based on MPI package found (without BML_ in front)

nicolasbock: Ah, I hadn't realized that. I agree with you to follow CMake's naming convention.

nicolasbock (approved):
Thanks!

nicolasbock (changes requested):
Sorry I clicked on approve too quickly. We should first find out why the MPI CI run failed before we merge this change.

nicolasbock (approved):
Thanks!

PR #437, opened 2021-01-14
Disable CircleCI

nicolasbock:
Signed-off-by: Nicolas Bock <nicolasbock@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/437)

PR #444, opened 2021-03-12
Cannon algorithm for distributed2d

jeanlucf22:
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/444)

### DIFF
--- a/src/C-interface/bml_utilities.c
+++ b/src/C-interface/bml_utilities.c
@@ -530,7 +530,7 @@ bml_write_bml_matrix(
     }
 }
 
-// compute largest int "i" such that i*i <= x
+// compute smallest int "i" such that i*i >= x
 int
 bml_sqrtint(
     const int x)
--- a/src/C-interface/bml_utilities.c
+++ b/src/C-interface/bml_utilities.c
@@ -529,3 +529,18 @@ bml_write_bml_matrix(
             break;
     }
 }
+
+// compute largest int "i" such that i*i <= x
+int
+bml_sqrtint(
+    const int x)
+{
+    int i = 1;
+    int result = 1;
+    while (result < x)
+    {
+        i++;
+        result = i * i;
+    }
+    return i;
+}
--- a/build.sh
+++ b/build.sh
@@ -115,7 +115,7 @@ set_defaults() {
     : ${BML_MAGMA:=no}
     : ${BML_CUSOLVER:=no}
     : ${BML_XSMM:=no}
-    : ${BML_ELLBLOCK_MEMPOOL:=yes}
+    : ${BML_ELLBLOCK_MEMPOOL:=no}
     : ${CUDA_TOOLKIT_ROOT_DIR:=}
     : ${INTEL_OPT:=no}
 }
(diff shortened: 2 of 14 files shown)


suemni (approved):

nicolasbock:
I think we should turn it off. See #437

jeanlucf22:
@nicolasbock ok. this PR is ready for merging too

nicolasbock:
I rebased the change

src/C-interface/bml_utilities.c
```diff
@@ -529,3 +529,18 @@ bml_write_bml_matrix(
             break;
     }
 }
+
+// compute largest int "i" such that i*i <= x
```
nicolasbock, line 533: Wouldn't
```
(int) floor(sqrt(x)))
```
Be a little simpler? Or are you concerned about performance and this implementation is faster?

nicolasbock (changes requested):

src/C-interface/bml_utilities.c
```diff
@@ -529,3 +529,18 @@ bml_write_bml_matrix(
...
+{
+    int i = 1;
+    int result = 1;
+    while (result < x)
+    {
+        i++;
+        result = i * i;
+    }
+    return i;
+}
```
nicolasbock, line 546:
```suggestion
{
    int i = 1;
    int result = 4;
    while (result < x)
    {
        i++;
        result = (i + 1) * (i + 1);
    }
    return i;
}
```

nicolasbock (changes requested):

src/C-interface/bml_utilities.c
jeanlucf22: Performance is not an issue since it is run only during setup. I just liked the idea of having an "int" specific function

src/C-interface/bml_utilities.c
jeanlucf22: @nicolasbock Not sure why you are suggesting that... Seems more complicated to me

nicolasbock: Sorry, I should have added a comment to explain :smile: The function as is does not work. For example `sqrtint(10) -> 4`.

src/C-interface/bml_utilities.c
nicolasbock: The type casting is a bit annoying, I agree. It might be more transparent when reading the code though. I am ok with either solution though. I was just interested in understanding why you had chosen this way of implementing it. Thanks!

src/C-interface/bml_utilities.c
jeanlucf22: I see. I guess the comment doesn't match the code, even though it works for all the square int (4, 9, 16,...) which is what we really care about

nicolasbock: Thanks! Yes, this makes more sense now :smile:

nicolasbock (approved):

nicolasbock:
Pushed a fix for my previous rebase attempt.

nicolasbock:
Thanks @jeanlucf22 !

PR #452, opened 2021-04-20
Add testing of remaining float types to MPI tests

nicolasbock:
So far we are testing only the `single_real` float type for the MPI
tests. This change adds the other float types.

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/452)


jeanlucf22 (commented):
I believe valgrind is off in this case even though the flag is on. Our cmake build turns it off if MPI is on for a good reason: it would fail due to MPI memory errors. So maybe we should remove the BML_VALGRIND flag here to avoid confusion? It would be nice to test with valgrind on, but I would not know how to do it with these MPI errors...

suemni (approved):

nicolasbock:
> I believe valgrind is off in this case even though the flag is on. Our cmake build turns it off if MPI is on for a good reason: it would fail due to MPI memory errors. So maybe we should remove the BML_VALGRIND flag here to avoid confusion? It would be nice to test with valgrind on, but I would not know how to do it with these MPI errors...

That's a good point @jeanlucf22 . I can disable VALGRIND until we figure out how to use it more usefully with MPI.

PR #454, opened 2021-04-20
Update CI image to Bionic

nicolasbock:
* Xenial is going out of regular maintenance mid-year 2021 [1]
* Xenial is EOL for LLVM since March 28 2021 [2]
* Xenial does not support >gcc-9 [3].

This change updates the CI image to using Bionic.

[1] https://ubuntu.com/about/release-cycle
[2] https://apt.llvm.org/
[3] https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test?field.series_filter=xenial

---
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/454)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -85,14 +85,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: lint
     container:
-      image: nicolasbock/bml:master
+      image: nicolasbock/bml:pr-454
     strategy:
       matrix:
         include:
-          - {CC: gcc-4.7, CXX: g++-4.7, FC: gfortran-4.7, BUILD_SHARED_LIBS: yes, BML_OPENMP: no,  BML_INTERNAL_BLAS: no}
-          - {CC: gcc-4.7, CXX: g++-4.7, FC: gfortran-4.7, BUILD_SHARED_LIBS: yes, BML_OPENMP: no,  BML_INTERNAL_BLAS: yes}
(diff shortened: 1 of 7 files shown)


.github/workflows/CI.yml
```diff
@@ -85,14 +85,14 @@ jobs:
     runs-on: ubuntu-latest
     needs: lint
     container:
-      image: nicolasbock/bml:master
+      image: nicolasbock/bml:pr-454
```
jeanlucf22, line 88: Is it correct?

nicolasbock: Yes. Let me explain:

The GitHub workflow has a step in it that builds the CI image and pushes that to Docker Hub. All other jobs in the CI workflow pull these images from Docker Hub and run their tests in them.

Ideally we would use the image built in a PR for the tests in that same PR so we can test the image and run our tests using that image. Unfortunately this is not possible with GitHub Actions. What we do instead is to tag the image with the branch so that PRs and the master branch are protected from each other. But that means that in order to use the image built in a PR we need to manually change the tag in the workflow file as I have done here.

Once we merge this PR I need to submit a follow up PR to change the tag back to `master`.

jeanlucf22: I trust you. I was just making sure it was correct.

jeanlucf22 (approved):

.github/workflows/CI.yml
nicolasbock: Thanks for pointing out that code change! That's the kind of change we need to watch out for because it can alter the CI process.

Maybe we should think about not using docker images so we can alter the test environment within a PR more easily. I'll have a look at that.

PR #457, opened 2021-04-21
Support OpenMP GPU offload on iris nodes at ANL

mewall:
This change is [Reviewable](https://reviewable.io/reviews/lanl/bml/457)

### DIFF
--- a/src/C-interface/ellpack/bml_add_ellpack_typed.c
+++ b/src/C-interface/ellpack/bml_add_ellpack_typed.c
@@ -57,7 +57,7 @@ void TYPED_FUNC(
     int rowMin = A_localRowMin[myRank];
     int rowMax = A_localRowMax[myRank];
 
-#if !(defined(__IBMC__) || defined(__ibmxl__))
+#if !(defined(__IBMC__) || defined(__ibmxl__) || (defined(USE_OMP_OFFLOAD) && defined(INTEL_SDK)))
     int ix[N], jx[N];
     REAL_T x[N];
 
@@ -66,12 +66,42 @@ void TYPED_FUNC(
     memset(x, 0.0, N * sizeof(REAL_T));
 #endif
 
+#if defined(USE_OMP_OFFLOAD) && defined(INTEL_SDK)
+    int num_chunks = MIN(OFFLOAD_NUM_CHUNKS, rowMax - rowMin + 1);
+
+    int all_ix[N * num_chunks], all_jx[N * num_chunks];
+    REAL_T all_x[N * num_chunks];
+
+    memset(all_ix, 0, N * num_chunks * sizeof(int));
+    memset(all_jx, 0, N * num_chunks * sizeof(int));
+    memset(all_x, 0.0, N * num_chunks * sizeof(REAL_T));
+
+#pragma omp target map(to:all_ix[0:N*num_chunks],all_jx[0:N*num_chunks],all_x[0:N*num_chunks])
+
+#endif
+
 #if defined (USE_OMP_OFFLOAD)
+#if defined(INTEL_SDK)
+#pragma omp teams distribute parallel for \
+    shared(rowMin, rowMax)                \
+    shared(A_index, A_value, A_nnz)       \
+    shared(B_index, B_value, B_nnz)
+    for (int chunk = 0; chunk < num_chunks; chunk++)
+    {
+        int *ix, *jx;
+        REAL_T *x;
+
+        ix = &all_ix[chunk * N];
+        jx = &all_jx[chunk * N];
+        x = &all_x[chunk * N];
+
+#else
 #pragma omp target teams distribute parallel for \
     shared(rowMin, rowMax)                \
     shared(A_index, A_value, A_nnz)       \
     shared(B_index, B_value, B_nnz)       \
     firstprivate(ix, jx, x)
+#endif
 #else
 #if defined(__IBMC__) || defined(__ibmxl__)
 #pragma omp parallel for \
@@ -86,9 +116,12 @@ void TYPED_FUNC(
     firstprivate(ix, jx, x)
 #endif
 #endif
+#if defined(USE_OMP_OFFLOAD) && defined(INTEL_SDK)
+    for (int i = rowMin + chunk; i < rowMax; i = i + num_chunks)
+#else
     for (int i = rowMin; i < rowMax; i++)
+#endif
     {
-
 #if defined(__IBMC__) || defined(__ibmxl__)
         int ix[N], jx[N];
         REAL_T x[N];
@@ -142,6 +175,9 @@ void TYPED_FUNC(
         }
         A_nnz[i] = ll;
     }
+#if defined(USE_OMP_OFFLOAD) && defined(INTEL_SDK)
+}
+#endif
 }
 
 /** Matrix addition.
--- a/src/C-interface/ellpack/bml_multiply_ellpack_typed.c
+++ b/src/C-interface/ellpack/bml_multiply_ellpack_typed.c
@@ -128,7 +128,7 @@ void *TYPED_FUNC(
     int rowMin = X_localRowMin[myRank];
     int rowMax = X_localRowMax[myRank];
 
-#if !(defined(__IBMC__) || defined(__ibmxl__))
+#if !(defined(__IBMC__) || defined(__ibmxl__) || (defined(USE_OMP_OFFLOAD) && defined(INTEL_SDK)))
     int ix[X_N], jx[X_N];
     REAL_T x[X_N];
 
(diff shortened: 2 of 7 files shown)


jeanlucf22:
@mewall How performance portable are these changes? To me it looks like strides are changed and may affect performance on some architectures...

mewall:
> @mewall How performance portable are these changes? To me it looks like strides are changed and may affect performance on some architectures...

 @jeanlucf22 Yes, I checked with Jamal (@jmohdyusof) about this in particular before going ahead with the pull request. We noted that the code will need to be adapted depending on architecture but that it would probably make sense to merge and address this later. The stride can affect GPU performance but the parallel pragma is on an outer loop so it's not clear to me whether the impact of changing from unit stride will be so large. Probably the memory access pattern is OK for CPU?

nicolasbock (changes requested):
Could you add more details to this change? It's not clear what Iris is and why these changes are necessary to build on Iris.

jeanlucf22:
> > @mewall How performance portable are these changes? To me it looks like strides are changed and may affect performance on some architectures...
> 
> @jeanlucf22 Yes, I checked with Jamal (@jmohdyusof) about this in particular before going ahead with the pull request. We noted that the code will need to be adapted depending on architecture but that it would probably make sense to merge and address this later. The stride can affect GPU performance but the parallel pragma is on an outer loop so it's not clear to me whether the impact of changing from unit stride will be so large. Probably the memory access pattern is OK for CPU?

I would at least allow nteams to be equal to 1 (which would make it match the original code). Maybe nteams=1 should be the default value? Maybe we should add a build option to set it to something else when the architecture requires it?

BTW, I have used an OpenMP outerloop (teams) in other codes for NVIDIA GPUs with good performance improvements, but I would write it in a way that the stride for the inner loop is 1. Does a longer stride helps on Iris?

suemni (approved):

jmohdyusof:
> BTW, I have used an OpenMP outerloop (teams) in other codes for NVIDIA GPUs with good performance improvements, but I would write it in a way that the stride for the inner loop is 1. Does a longer stride helps on Iris? 

Do you have examples?

jeanlucf22:
See
https://github.com/jefflarkin/GTC16-S6510/blob/master/c99/05-target-teams-seperated-schedule/laplace2d.c
(not my code, but the model I followed)

mewall:
@jeanlucf22 @jmohdyusof 

I rewrote the code to make it more comparable to what's in master, using an omp parallel pragma on the outer t loop, and put a timer for the bml_multiply call in the multiply test. Following are times for N=1000 on a darwin V100 node:

unit stride (master): 

TIMING: bml_multiply = 20.401 secs

nteams stride:

TIMING: bml_multiply = 20.8417 secs

These timings are pretty consistent with what nvprof is reporting. 

Note: nteams essentially acts as an upper limit for the number of concurrent threads, so we don't really want to set it to 1.

I'll put some effort into the optimization next. Also I realized that there are other functions that might not exercised by the tests that will probably need changing to remove the firstprivate clause, e.g. bml_add_identity_ellpack()...maybe we should add some more tests?

mewall:
> Could you add more details to this change? It's not clear what Iris is and why these changes are necessary to build on Iris.

Yes. However, we decided today to hold this pull request open while we do some more work on optimization. I'll address this comment when the code is ready and more details are known.

nicolasbock:
> > Could you add more details to this change? It's not clear what Iris is and why these changes are necessary to build on Iris.
> 
> Yes. However, we decided today to hold this pull request open while we do some more work on optimization. I'll address this comment when the code is ready and more details are known.

Thanks @mewall !

jeanlucf22:
@mewall do you mean 20 usec? also, I would use larger matrices, maybe N=5000. 1000 is small for a GPU.

cnegre:
20 seconds for 1000×1000 matrix seems slow. I would expect 0.2 seconds

mewall:
The number I got was 20 secs for a 1000x1000 secs, using the code in the master branch.

From: Christian F. A. Negre ***@***.***>
Sent: Friday, April 23, 2021 4:28:28 AM
To: lanl/bml
Cc: Wall, Michael E; Mention
Subject: [EXTERNAL] Re: [lanl/bml] Modifications and scripts to build on iris (#457)


20 seconds for 1000×1000 matrix seems slow. I would expect 0.2 seconds

mewall:
I should clarify that I used N = 1000 but didn't specify M for this timing. When I specify M = 100 the time decreases to ~200 ms.


I did some preliminary experimentation with increasing the threading for GPUs, but it seems nontrivial to merge the results from different threads if we also parallelize the jp loop. I might be wandering down the same path Jamal already explored in coming up with his merge method, might it be better to see if we can implement that method using omp offload than for me to optimize what's there?

src/C-interface/ellpack/bml_add_ellpack_typed.c
```diff
@@ -57,90 +57,87 @@ void TYPED_FUNC(
...
-    shared(B_index, B_value, B_nnz)       \
-    firstprivate(ix, jx, x)
-#else
-#if defined(__IBMC__) || defined(__ibmxl__)
-#pragma omp parallel for \
-    shared(rowMin, rowMax)                \
-    shared(A_index, A_value, A_nnz)       \
-    shared(B_index, B_value, B_nnz)
-#else
+#pragma omp target map(to:all_ix[0:N*nteams], \
```
jeanlucf22, line 70: where did the "teams distribute parallel for" go? We should keep it

src/C-interface/ellpack/bml_multiply_ellpack_typed.c
```diff
@@ -128,116 +128,107 @@ void *TYPED_FUNC(
...
-#if defined(__IBMC__) || defined(__ibmxl__)
-#pragma omp parallel for                               \
-    shared(X_N, X_M, X_index, X_nnz, X_value)  \
-    shared(X2_N, X2_M, X2_index, X2_nnz, X2_value)     \
-    shared(rowMin, rowMax)                             \
-    reduction(+: traceX, traceX2)
-#else
-#ifndef CRAY_SDK
-#pragma vector aligned
+#pragma omp target map(to:all_ix[0:X_N*nteams],all_jx[0:X_N*nteams],all_x[0:X_N*nteams])
```
jeanlucf22, line 141: same here

src/C-interface/ellpack/bml_multiply_ellpack_typed.c
```diff
@@ -289,100 +280,91 @@ void TYPED_FUNC(
...
 #if defined (USE_OMP_OFFLOAD)
-#pragma omp target teams distribute parallel for \
-    shared(A_N, A_M, A_nnz, A_index, A_value)  \
-    shared(A_localRowMin, A_localRowMax)       \
-    shared(B_N, B_M, B_nnz, B_index, B_value)  \
-    shared(C_N, C_M, C_nnz, C_index, C_value)  \
-    firstprivate(ix, jx, x)
-#else
-#if defined(__IBMC__) || defined(__ibmxl__)
+#pragma omp target map(to:all_ix[0:C_N*nteams],all_jx[0:C_N*nteams],all_x[0:C_N*nteams])
```
jeanlucf22, line 293: and here

jeanlucf22 (changes requested):

src/C-interface/ellpack/bml_multiply_ellpack_typed.c
mewall: Apologies, this isn't ready for review. I'm just capturing the state before I incorporate the changes discussed in today's meeting about the Iris build, which will be fairly extensive.

Let me know if I should close the pull request or keep it open while I revise.

nicolasbock: You could convert the PR to a draft while working on it. I would prefer that over closing it altogether.

mewall: @nicolasbock @jeanlucf22 Done, thanks.

mewall: @nicolasbock @jeanlucf22 This is now ready for review.

mewall:
> > > @mewall How performance portable are these changes? To me it looks like strides are changed and may affect performance on some architectures...
> > 
> > 
> > @jeanlucf22 Yes, I checked with Jamal (@jmohdyusof) about this in particular before going ahead with the pull request. We noted that the code will need to be adapted depending on architecture but that it would probably make sense to merge and address this later. The stride can affect GPU performance but the parallel pragma is on an outer loop so it's not clear to me whether the impact of changing from unit stride will be so large. Probably the memory access pattern is OK for CPU?
> 
> I would at least allow nteams to be equal to 1 (which would make it match the original code). Maybe nteams=1 should be the default value? Maybe we should add a build option to set it to something else when the architecture requires it?
> 
> BTW, I have used an OpenMP outerloop (teams) in other codes for NVIDIA GPUs with good performance improvements, but I would write it in a way that the stride for the inner loop is 1. Does a longer stride helps on Iris?



> > > @mewall How performance portable are these changes? To me it looks like strides are changed and may affect performance on some architectures...
> > 
> > 
> > @jeanlucf22 Yes, I checked with Jamal (@jmohdyusof) about this in particular before going ahead with the pull request. We noted that the code will need to be adapted depending on architecture but that it would probably make sense to merge and address this later. The stride can affect GPU performance but the parallel pragma is on an outer loop so it's not clear to me whether the impact of changing from unit stride will be so large. Probably the memory access pattern is OK for CPU?
> 
> I would at least allow nteams to be equal to 1 (which would make it match the original code). Maybe nteams=1 should be the default value? Maybe we should add a build option to set it to something else when the architecture requires it?
> 
> BTW, I have used an OpenMP outerloop (teams) in other codes for NVIDIA GPUs with good performance improvements, but I would write it in a way that the stride for the inner loop is 1. Does a longer stride helps on Iris?

@jeanlucf22 @jmohdyusof FYI I addressed this in #457 by making the Iris changes specific to USE_OMP_OFFLOAD and INTEL_SDK being defined. There should be no impact now on the usual code path.

mewall:
@cnegre This might address issue #242 -- new macro INTEL_SDK was introduced for intel builds.

nicolasbock (approved):
This looks good to me, but I haven't tested the code.  Note that we are not testing in CI with `OMP_OFFLOAD` for example and any regression in those code paths would go unnoticed.

jeanlucf22 (approved):

Issue #404, opened 2020-08-08
Remove --coverage flag for default Debug build


jeanlucf22:
The flag --coverage is added to the GNU_Fortran_FLAGS_DEBUG. When building BML in Debug mode, this flag is always ON. There is no clean way to turn it off.

This flag should be added only when one chooses that option.

PR #473, opened 2021-04-27
Add extra coverage job

nicolasbock:


jeanlucf22 (commented):
I thought that we were not testing for coverage anymore. @nicolasbock could you clarify what that change does?

nicolasbock:
You are right, this was a not very thought out PR right before a meeting. I had just removed the default coverage flag for #404 and thought I could squeeze in just one more PR :)

What do you think if I turn this into a PR that completely removes the flag from the build system?

nicolasbock:
Abandoning

PR #482, opened 2021-05-06
Print threshold

jmohdyusof:
Implementing some test-specific printing thresholds to limit output when testing on large matrices.


suemni (approved):

jmohdyusof:
There are some weird errors in the build checks, attempting to fix.

nicolasbock (approved):
Thanks.  That should work. I was hoping we could address this without having to hardcode the `PRINT_THRESHOLD` but it's not clear to me how we would do that.

jmohdyusof:
Yeah, without passing around yet another argument I don't see a way of making print_threshold settable at runtime

jmohdyusof:
@nicolasbock Do you understand what is causing the test error? It appears that the print_bml_matix called from the multiply test cannot infer the type of C? multiply_matrix_typed, line 169

tests/C-tests/multiply_matrix_typed.c
```diff
@@ -154,20 +160,28 @@ int TYPED_FUNC(
...
     REAL_T *D_dense = bml_export_to_dense(C, dense_row_major);
 
-    LOG_INFO("bml_multiply\n");
+    if (bml_getMyRank() == 0)
+        LOG_INFO("bml_multiply\n");
     bml_multiply(A, B, C, alpha, beta, threshold);
+    if (bml_getMyRank() == 0)
+    {
+        LOG_INFO("C = %f * A + %f * B [0: %d][0: %d]\n", alpha, beta, N, N);
+        bml_print_bml_matrix(C, 0, max_row, 0, max_col);
```
jeanlucf22: I think this is the issue: bml_print_bml_matrix() is not implemented yet for the 'distributed2d' format. I should have it soon...

jmohdyusof:
OK, I guess this is why everything uses export_to_dense and then print_dense?

tests/C-tests/multiply_matrix_typed.c
nicolasbock: :+1:

nicolasbock:
> Yeah, without passing around yet another argument I don't see a way of making print_threshold settable at runtime

The only other approach I can see right now is to use an environment variable. Basically not restrict output by default but check whether this environment variable is set. If it is then use its value.

jmohdyusof:
> > Yeah, without passing around yet another argument I don't see a way of making print_threshold settable at runtime
> 
> The only other approach I can see right now is to use an environment variable. Basically not restrict output by default but check whether this environment variable is set. If it is then use its value.

If someone else wants to implement that I am fine with it.

jeanlucf22:
> OK, I guess this is why everything uses export_to_dense and then print_dense?

No, most prints were like this already. But using the bml_print function would be cleaner to use. Maybe something to change once I have added the option for 'distributed2d'...

jmohdyusof:
> > OK, I guess this is why everything uses export_to_dense and then print_dense?
> 
> No, most prints were like this already. But using the bml_print function would be cleaner to use. Maybe something to change once I have added the option for 'distributed2d'...

In fact, all the print_bml_matrix code does under the covers is export_to_dense and then print_dense, so it really is just that the code is falling through the case selection loop.

jeanlucf22 (approved):

nicolasbock:
> > > Yeah, without passing around yet another argument I don't see a way of making print_threshold settable at runtime
> > 
> > 
> > The only other approach I can see right now is to use an environment variable. Basically not restrict output by default but check whether this environment variable is set. If it is then use its value.
> 
> If someone else wants to implement that I am fine with it.

That sounds good. Let's merge this now and then potentially improve the code later.

PR #483, opened 2021-05-06
Distributed2d diagonalization and norm functions

jeanlucf22:
Diagonalization requires ScaLAPACK.
ScaLAPACK library should be specified using
export EXTRA_LINK_FLAGS=...
since cmake build has currently not capability to find it.

PR #484, opened 2021-05-06
Enable tests with scalapack

nicolasbock:
This pull request should be merged after lanl/bml#483 is merged.


suemni (approved):

jeanlucf22 (approved):
Great, thanks @nicolasbock !

.github/workflows/CI.yml
```diff
@@ -119,6 +119,7 @@ jobs:
           - {CC: mpicc,   CXX: mpic++,  FC: mpifort,      BUILD_SHARED_LIBS: no,  BML_OPENMP: no,  BML_INTERNAL_BLAS: no, BML_MPI: yes, TESTING_EXTRA_ARGS: "-R MPI-C-.*-double_real"}
           - {CC: mpicc,   CXX: mpic++,  FC: mpifort,      BUILD_SHARED_LIBS: no,  BML_OPENMP: no,  BML_INTERNAL_BLAS: no, BML_MPI: yes, TESTING_EXTRA_ARGS: "-R MPI-C-.*-single_complex"}
           - {CC: mpicc,   CXX: mpic++,  FC: mpifort,      BUILD_SHARED_LIBS: no,  BML_OPENMP: no,  BML_INTERNAL_BLAS: no, BML_MPI: yes, TESTING_EXTRA_ARGS: "-R MPI-C-.*-double_complex"}
+          - {CC: mpicc,   CXX: mpic++,  FC: mpifort,      BUILD_SHARED_LIBS: no,  BML_OPENMP: no,  BML_INTERNAL_BLAS: no, BML_MPI: yes, TESTING_EXTRA_ARGS: "-R MPI-C-.*-single_real", EXTRA_LINK_FLAGS: -lscalapack-mpich, BML_SCALAPACK: yes}
```
jeanlucf22, line 122: We should add Scalapack to the 4 test runs above instead of adding a new one

nicolasbock: done

nicolasbock:
> Looks like the failures are due to the fact that MPI tests are running with 1 task only instead of 4. The number of tasks (4) is hard coded in tests/C-tests/CMakeLists.txt

Thanks for the hint! I removed the hardcoded `4` in https://github.com/lanl/bml/pull/484/files#diff-2d6e142a15f8a12f75f4a96846acb9fe0b8cdbdeb6c7ce20d9ade05ca35f73d5L117

jeanlucf22:
I did not mean that a hard coded 4 was a problem. But the tests are written assuming a number of tasks>1, and equal to the square of an int. What it looked like was that the MPI tests were not using that 4, but were using 1 instead...

nicolasbock:
Ah ok, let''s try this again then :smile:

jeanlucf22:
I have been using MPIEXEC_NUMPROCS_FLAG for something like "-np", not the number of tasks itself. You seem to be overriding "-np" with 4 here...

nicolasbock:
I cleaned up the flags a bit. We can't use one variable for two arguments such as `-np 4` because the shell will pass it as one argument to `mpirun` as opposed to two.

nicolasbock:
The `mpi_sendrecv` tests are failing. I am a little confused why they claim that they only get 1 rank (https://github.com/lanl/bml/pull/484/checks?check_run_id=2535554023#step:4:2585). This is the case also for other MPI tests but those don't fail on 1 rank. I can't find anything obviously wrong with the `MPI_Init` call. 

Running

```
mpiexec -np 4 echo "process"
```

in the CI container results in 4 lines of output as expected, suggesting that the container and its `mpich` installation work. Maybe the `mpicc` compiler in the container is not working?

jeanlucf22:
I am confused too, because these tests were passing before... The only difference is linking with ScaLapack now, right?

nicolasbock:
Yes that's the only change. In local testing I was able to get my test program to work with OpenMPI instead of MPICH. Maybe we are missing some environment variable for MPICH? I kind of vaguely remember that there was something needed there.

nicolasbock:
`BML_USE_SCALAPACK` is only used in [bml_diagonalize_distributed2d_typed.c](https://github.com/lanl/bml/blob/master/src/C-interface/distributed2d/bml_diagonalize_distributed2d_typed.c#L150), isn't it?

Scalapack introduces an alternative code path in this code. Shouldn't we test with and without scalapack to test both paths?

nicolasbock:
Never mind my comment about the two code paths. The second one without scalapack is very short 😉

nicolasbock:
`OpenMPI` seems to work actually 🍾

PR #490, opened 2021-05-20
A few more functions for distributed2d

jeanlucf22:


nicolasbock (approved):

nicolasbock (changes requested):
Could you look into why [this](https://github.com/lanl/bml/pull/490/checks?check_run_id=2630610180#step:4:2778) is failing?

suemni (approved):

jeanlucf22:
@nicolasbock There was actually a bug in the diagonalization function for the distributed2d format that we were not catching before. Just fixed it, and fixed the test too.

nicolasbock:
> @nicolasbock There was actually a bug in the diagonalization function for the distributed2d format that we were not catching before. Just fixed it, and fixed the test too.

Nice!

There is still a memory leak in one of the tests: https://github.com/lanl/bml/pull/490/checks?check_run_id=2643057870#step:4:4114

Could you have a look?

Thanks!

nicolasbock (approved):
Thanks @jeanlucf22 !

PR #516, opened 2021-07-11
Version bump to v2.0.1

nicolasbock:


jeanlucf22 (commented):
Should we make it 2.1? There has been many features added since 2.0...

nicolasbock:
I was going to release `v2.0.1` as a maintenance release to include the fix for the version. Currently we don't don't have a branch (or tag) that includes it. But yes, we should then also release `v2.1.0` from `master` to include all of the other improvements. I'll add a pull request for that separetely.

jeanlucf22 (approved):

PR #522, opened 2021-07-11
Add mixed clang/gcc tests

nicolasbock:

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -86,6 +86,8 @@ jobs:
             TEST_SCRIPT: ./scripts/ci-clang-11-gfortran-11-C-single-real.sh
           - JOBNAME: clang-11 / gfortran-11 C double real
             TEST_SCRIPT: ./scripts/ci-clang-11-gfortran-11-C-double-real.sh
+          - JOBNAME: clang-11 / gfortran-11 C single complex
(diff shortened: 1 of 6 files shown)


nicolasbock:
@awvwgk While I figure out how to build on OS X I added a mixed clang/gcc test to Linux.

jeanlucf22 (commented):
why 'gcc-9' in the names of these new scripts? Should it be 'gfortran9'?
Also, could it be using flang instead of gfortran?

nicolasbock:
> why 'gcc-9' in the names of these new scripts? Should it be 'gfortran9'?

Yes, I can definitely change it. I had meant `gcc` more as the GNU compiler collection generically and not specifically as their C compiler. But the naming should as clear as possible and `gfortran-9` would indicate perhaps more clearly that it's only the Fortran part that is using GCC.

> Also, could it be using flang instead of gfortran?

Yes, that might be more sensible. I'll change that as well.

Thanks!

nicolasbock:
> > Also, could it be using flang instead of gfortran?
> 
> Yes, that might be more sensible. I'll change that as well.

So far I haven't found any `flang` packages. Are you using `flang`? How do you install it if you do?

jeanlucf22:
I don't write much fortran myself, so I don't know. But I am ok with clang + gfortran if they work well together.

cnegre (approved):

Issue #515, opened 2021-07-10
Build failure on OSX


awvwgk:
**Describe the bug**

Building the Fortran bindings for BML 1.3.1 currently fails with a linking error on OSX due to unresolved symbols:

```
[  7%] Linking Fortran shared library libbml_fortran.dylib
ld: warning: -pie being ignored. It is only used when linking a main executable
Undefined symbols for architecture x86_64:
  "_bml_add", referenced from:
      ___bml_add_m_MOD_add_two_deprecated in bml_add_m.F90.o
      ___bml_add_m_MOD_add_two in bml_add_m.F90.o
     (maybe you meant: ___bml_add_m_MOD_bml_add_norm, ___bml_add_m_MOD_add_identity_one , ___bml_add_m_MOD_bml_scale_add_identity , ___bml_add_m_MOD_add_two , ___bml_add_m_MOD_add_two_deprecated )
  "_bml_add_identity", referenced from:
      ___bml_add_m_MOD_add_identity_one in bml_add_m.F90.o
...
  "_bml_sum_squares_submatrix", referenced from:
      ___bml_norm_m_MOD_bml_sum_squares_submatrix in bml_norm_m.F90.o
     (maybe you meant: ___bml_norm_m_MOD_bml_sum_squares_submatrix)
  "_bml_threshold", referenced from:
      ___bml_threshold_m_MOD_bml_threshold in bml_threshold_m.F90.o
     (maybe you meant: ___bml_threshold_m_MOD_bml_threshold)
  "_bml_trace", referenced from:
      ___bml_trace_m_MOD_bml_trace in bml_trace_m.F90.o
     (maybe you meant: ___bml_trace_m_MOD_bml_trace, ___bml_trace_m_MOD_bml_tracemult )
  "_bml_traceMult", referenced from:
      ___bml_trace_m_MOD_bml_tracemult in bml_trace_m.F90.o
  "_bml_transpose_new", referenced from:
      ___bml_transpose_m_MOD_bml_transpose in bml_transpose_m.F90.o
  "_bml_transpose_triangle", referenced from:
      ___bml_transpose_triangle_m_MOD_bml_transpose_triangle in bml_transpose_triangle_m.F90.o
     (maybe you meant: ___bml_transpose_triangle_m_MOD_bml_transpose_triangle)
  "_bml_update_domain", referenced from:
      ___bml_allocate_m_MOD_bml_update_domain in bml_allocate_m.F90.o
[...]

PR #523, opened 2021-07-11
Test workflow for OSX

awvwgk:
Add a test workflow for OSX with GCC 5 / GFortran 5, currently failing.

- Link error can be fixed by declaring test library as `STATIC`, maybe better to add `target_link_libraries` with `bml` target (see #515)
- Segmentation fault when built with `BML_INTERNAL_BLAS=yes` (gcc-5 test 2 & gcc-5 test 4)
- Regular expression mismatch for `test-backstrace` (gcc-5 test 1 & gcc-5 test 3)

Opening this PR for further feedback, the `test-backtrace` regex can be fixed quite easily, not sure what is causing the segmentation fault, though.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -91,3 +91,30 @@ jobs:
         env:
           TEST_SCRIPT: ${{ matrix.TEST_SCRIPT || env.TEST_SCRIPT }}
         run: ${TEST_SCRIPT} testing
+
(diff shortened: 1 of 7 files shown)


nicolasbock:
Thanks @awvwgk for submitting the additional jobs. I am going to have a closer look at the logs.

cnegre (approved):

PR #534, opened 2021-07-25
added sparse matrix example

finkeljos:
Added a .c file to /bml/examples/sparse-matrix/ that generates a sparse Hamiltonian matrix using the example Hamiltonian shown in https://doi.org/10.1021/acs.jctc.1c00057

### DIFF
--- a/examples/sparse-matrix/gen-sparse.c
+++ b/examples/sparse-matrix/gen-sparse.c
@@ -9,7 +9,8 @@ Generate a sparse N x N Hamiltonian matrix given the matrix dimension, N, and th
 Output to terminal given in matrix market format.
 
 Usage: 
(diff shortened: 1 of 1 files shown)


cnegre (approved):

cnegre:
@finkeljos Try to fix the indentation with the indent.sh script

nicolasbock (changes requested):
As @cnegre already said, indentation is incorrect.

jeanlucf22 (changes requested):
The header line for the matrix market format seems to be missing in the created file. I am not sure it is always required, but it is good practice to have a short description of the type of matrix in the file

jeanlucf22 (approved):

nicolasbock (approved):

PR #536, opened 2021-08-10
Fix copy and setters

oseikuffuor1:
This PR:
1. Fixes an issue with bml_copy where we check for storage size using the initial column dimension. For CSR, this check is done internally, since individual rows need not have the same size.
2. Modifies bml_setters_csr to avoid allocating space for a dense row when setting sparse row entries from a dense array.

### DIFF
--- a/src/C-interface/csr/bml_copy_csr_typed.c
+++ b/src/C-interface/csr/bml_copy_csr_typed.c
@@ -76,51 +76,6 @@ bml_matrix_csr_t *TYPED_FUNC(
     return B;
 }
 
-/** Copy an csr matrix - result is a new matrix.
- * Also allocates hash table and lvarsgid data
- *
- *  \ingroup copy_group
- *
- *  \param A The matrix to be copied
- *  \return A copy of matrix A.
- */
-/*
-bml_matrix_csr_t *TYPED_FUNC(
-    bml_copy_csr_new) (
-    const bml_matrix_csr_t * A)
-{
-    bml_matrix_csr_t *B =
-        TYPED_FUNC(bml_noinit_matrix_csr) (A->N_, A->NZMAX_,
-                                            A->TOTNNZ_, A->distribution_mode);
-
-    int N = A->N_;
-    csr_row_index_hash_t *table;
-
-    // copy lvarsgid_
-    memcpy(B->lvarsgid_, A->lvarsgid_, sizeof(int) * N);
-    // allocate hash table. Cannot copy table due to pointer incompatibilities
-    table = (csr_row_index_hash_t *)malloc(sizeof(csr_row_index_hash_t) * N);
-    for(int i=0; i<N; i++)
-    {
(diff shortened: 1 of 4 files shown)


src/C-interface/csr/bml_setters_csr_typed.c
```diff
@@ -97,7 +97,7 @@ void TYPED_FUNC(
  *
  *
  */
-void TYPED_FUNC(
+/*void TYPED_FUNC(
```
jeanlucf22, line 100: Is there a reason to keep that old function commented out?

oseikuffuor1: Nope. Just forgot to clean that up. I'll do that now.

src/C-interface/csr/bml_copy_csr_typed.c
```diff
@@ -136,7 +136,11 @@ void TYPED_FUNC(
 {
     const int NNZ = arow->NNZ_;
     // Check size for data
-    assert(brow->alloc_size_ >= NNZ);
+//    assert(brow->alloc_size_ >= NNZ);
```
nicolasbock, line 139: Could you also remove this line? I don't see a reason to keep it.

nicolasbock (changes requested):

src/C-interface/csr/bml_copy_csr_typed.c
oseikuffuor1: Got it.

jeanlucf22 (approved):

cnegre (approved):

nicolasbock:
@oseikuffuor1 The pull request description exists only on GitHub and is not part of the repository. For future forensics it would be useful to have as much information on a commit as possible which includes your description. In addition, multiple trivial commits such as indentation fixes are distracting when reading the code history and should be avoided.

Could you copy your pull request description into the commit comment and squash the commits into one?

oseikuffuor1:
@nicolasbock I limited the text in my commit message since I believe there used to be a limit to the number of characters we could use for the commit message to BML. Has this requirement changed?

nicolasbock:
The only limit is that the first line (the subject) is no longer than 50 characters. Otherwise there is no length restriction. Some additional things to consider:

1. Separate subject from body with a blank line
2. Limit the subject line to 50 characters
3. Capitalize the subject line
4. Do not end the subject line with a period
5. Use the imperative mood in the subject line
6. Wrap the body at 72 characters
7. Use the body to explain what and why vs. how

Hope this helps.

oseikuffuor1:
Thanks @nicolasbock. This helps. I will update the commit message and do the squash commit for this PR as you suggest.

oseikuffuor1:
@nicolasbock I made the suggested changes in the latest commit [here](https://github.com/lanl/bml/pull/536/commits/3b6e21dacc56399258a68e53577b147779b17e42) . Thanks

PR #537, opened 2021-08-11
Fix paths in tag update script

nicolasbock:
Since the script was moved into the `scripts` directory, the `basedir`
needs to be updated to include files starting from the repository root
directory.


jeanlucf22 (approved):

nicolasbock:
Thanks!

PR #543, opened 2021-10-06
Change the way MPI datatype is built for dense case

jeanlucf22:
I have to admit I don't fully understand why this change makes a difference, but it does on some platforms at least where the test suite fails without that fix.
Note that it is already done this way for other matrix formats.

### DIFF
--- a/src/C-interface/dense/bml_parallel_dense_typed.c
+++ b/src/C-interface/dense/bml_parallel_dense_typed.c
@@ -64,11 +64,10 @@ void TYPED_FUNC(
     MPI_Datatype dtype[1];
     dtype[0] = MPI_T;
 
-    int blength[1];
-    blength[0] = A->N * A->N;
-    MPI_Aint displ[0];
+    int blength = A->N * A->N;
+    MPI_Aint displ[1];
     displ[0] = addr0 - baseaddr;
-    int mpiret = MPI_Type_create_struct(1, blength, displ, dtype, newtype);
+    int mpiret = MPI_Type_create_struct(1, &blength, displ, dtype, newtype);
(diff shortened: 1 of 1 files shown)


suemni (approved):

nicolasbock:
Unfortunately now the CI tests are failing for dense.

src/C-interface/dense/bml_parallel_dense_typed.c
```diff
@@ -64,10 +64,11 @@ void TYPED_FUNC(
     MPI_Datatype dtype[1];
     dtype[0] = MPI_T;
 
-    int blength = A->N * A->N;
+    int blength[1];
+    blength[0] = A->N * A->N;
     MPI_Aint displ[0];
     displ[0] = addr0 - baseaddr;
-    int mpiret = MPI_Type_create_struct(1, &blength, displ, dtype, newtype);
+    int mpiret = MPI_Type_create_struct(1, blength, displ, dtype, newtype);
```
nicolasbock, line 71: The code looks correct though, although I don't understand why it wasn't working before. `count = 1` should work with `int&`.

src/C-interface/dense/bml_parallel_dense_typed.c
```diff
@@ -65,7 +65,7 @@ void TYPED_FUNC(
     dtype[0] = MPI_T;
 
     int blength = A->N * A->N;
-    MPI_Aint displ[0];
+    MPI_Aint displ[1];
```
nicolasbock, line 68: Nice catch! I hadn't noticed that 😞

nicolasbock (approved):

PR #544, opened 2021-10-06
Add CUDA libraries when linking if CUDA is found by cmake

mewall:
Fixes a bug where the link step fails when building on LANL Darwin cluster, due to missing CUDA libs in the link line

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,6 +238,7 @@ if(BML_CUDA OR BML_CUSOLVER)
 endif()
 if (CUDA_FOUND)
    message("CUDA CUBLAS libraries: ${CUDA_CUBLAS_LIBRARIES}")
+   message("CUDA libraries: ${CUDA_LIBRARIES}")
    include_directories(${CUDA_INCLUDE_DIRS})
 endif()
 
@@ -272,6 +273,14 @@ if (MAGMA_FOUND)
     message(STATUS "Will use cuSOLVER")
     add_definitions(-DBML_USE_CUSOLVER)
   endif()
+endif()
+
+if (CUBLAS_FOUND)
+  list(APPEND LINK_LIBRARIES ${CUBLAS_LIBRARIES})
+endif()
+
+if (CUDA_FOUND)
+  list(APPEND LINK_LIBRARIES ${CUDA_LIBRARIES})
 endif()
 
 set(BML_SCALAPACK FALSE CACHE BOOL "Whether to use ScaLAPACK library")


nicolasbock (approved):
LGTM

jmohdyusof:
As we discussed on the call I am not sure which, if any, of these flags is actually set.

nicolasbock:
You mean `CUBLAS_LIBRARIES` and `CUDA_LIBRARIES`? If `CUDA_FOUND` or `CUBLAS_FOUND` is `true` shouldn't those be set?

jmohdyusof:
The current code assumes CUDA_CUBLAS_LIBRARIES is the flag that is set. From the CMakelists.txt there is no way to tell what flags are actually being set, and I don't know myself. We are also discussing that we need to generalize BML_OFFLOAD to be able to check for the analogous AMD and Intel, etc libraries depending on the platform.

mewall:
I just removed the CUBLAS_LIBRARIES block as that wasn't needed.


For some reason the CUDA_LIBRARIES weren't added when I built on Darwin, so this PR adds them explicitly, which makes the build work.


Note that according to the following page FindCUDA is deprecated, so we should probably change the way that's done anyway:


https://cmake.org/cmake/help/latest/module/FindCUDA.html



Sent: Wednesday, October 6, 2021 11:42:17 AM
To: lanl/bml
Cc: Wall, Michael E; Author
Subject: [EXTERNAL] Re: [lanl/bml] Add CUDA libraries when linking if CUDA is found by cmake (#544)


You mean CUBLAS_LIBRARIES and CUDA_LIBRARIES? If CUDA_FOUND or CUBLAS_FOUND is true shouldn't those be set?

mewall:
I'm about to submit another PR that combines the two CUDA_FOUND blocks into one, and eliminates the CUBLAS_FOUND block.


I'll add a couple of issues for the revisions to the CUDA detection and the BML_OFFLOAD parts.

From: Jamal Mohd-Yusof ***@***.***>
Sent: Wednesday, October 6, 2021 11:45:15 AM
To: lanl/bml
Cc: Wall, Michael E; Author
Subject: [EXTERNAL] Re: [lanl/bml] Add CUDA libraries when linking if CUDA is found by cmake (#544)


The current code assumes CUDA_CUBLAS_LIBRARIES is the flag that is set. From the CMakelists.txt there is no way to tell what flags are actually being set, and I don't know myself. We are also discussing that we need to generalize BML_OFFLOAD to be able to check for the analogous AMD and Intel, etc libraries depending on the platform.

PR #545, opened 2021-10-06
Add CUDA libraries when linking if CUDA is found by cmake

mewall:


jmohdyusof:
Still curious if manually appending to the list or using a function like include_directories() is preferred.

mewall:
I just checked, simply adding CUDA as a language doesn't add the CUDA include path at compile time, at least in my hands.


I can't work on this more now but I did create an issue for it already. Maybe accept the latest PR and work on this later? The new PR just changes one line of code in CMakeLists.txt compared to commit 27295f941c8f.

Sent: Wednesday, October 6, 2021 12:24:35 PM
To: lanl/bml
Cc: Wall, Michael E; Mention
Subject: [EXTERNAL] Re: [lanl/bml] Add CUDA libraries when linking if CUDA is found by cmake (#545)


@nicolasbock requested changes on this pull request.

It is cleaner to just remove the FindCUDA module and all of the logic that goes with it and replace it with listing CUDA as a language at the top.

jmohdyusof:
@nicolasbock What I was trying to say was that maybe there should be a link_directories() function that adds to the link_libraries. It seems incongruous to use two different methods to do very similar things (since the include_directories should also append to some list of include paths)

cnegre (approved):

PR #548, opened 2021-10-11
set_row distributed2d

jeanlucf22:


src/C-interface/bml_getters.c
```diff
@@ -6,6 +6,7 @@
 #include "ellsort/bml_getters_ellsort.h"
 #include "ellblock/bml_getters_ellblock.h"
 #include "csr/bml_getters_csr.h"
+#include "distributed2d/bml_getters_distributed2d.h"
```
nicolasbock, line 9: If you unconditionally include this here then it will pull in `bml_types_distributed2d.h` which will in turn require MPI support. Maybe try to wrap the `bml_matrix_distributed2d_t` in `bml_types_distributed2d.h` with an `ifdef`.

nicolasbock (changes requested):

nicolasbock:
The tests are working again @jeanlucf22 . Please have a look at my comment and the CI failures for the non-MPI tests.

src/C-interface/bml_getters.c
nicolasbock: 👍🏾

PR #551, opened 2021-10-13
Add bml_get_deep_type function

jeanlucf22:


src/C-interface/bml_introspection.c
```diff
@@ -35,6 +35,24 @@ bml_get_type(
 
 }
```
nicolasbock, line 37: Please add documentation to this new function. If I understand it correctly then it's used to inspect the matrix types underneath the distributed2d type.

nicolasbock (approved):
Thanks @jeanlucf22

PR #558, opened 2021-11-11
Use irecv in distributed transpose

jeanlucf22:
This fixes a deadlock seen for matrices larger than those
in test suite


nicolasbock (approved):
LGTM

PR #560, opened 2021-11-14
Fix issues with ScaLapack call

jeanlucf22:
Size of work array had to be fixed. In addition, that size had to be increased
to work around possible bug in ScaLapack.
Failure was not seen for matrix size used in test suite, but only for larger
matrices.

### DIFF
--- a/src/C-interface/distributed2d/bml_diagonalize_distributed2d_typed.c
+++ b/src/C-interface/distributed2d/bml_diagonalize_distributed2d_typed.c
@@ -202,15 +202,15 @@ void TYPED_FUNC(
     assert(atmp != NULL);
 
     int ione = 1;
-#if defined(SINGLE_REAL) || defined(DOUBLE_REAL)
-    int lwork = MAX(1 + 6 * m + 2 * mb * mb,
-                    3 * m + MAX(mb * (mb + 1), 3 * mb)) + 2 * m;
-#else
     int izero = 0;
-    int np0 = NUMROC(&m, &mb, &izero, &izero, &np_rows);
+    int np0 = NUMROC(&m, &mb, &my_prow, &izero, &np_rows);
+    int nq0 = NUMROC(&m, &mb, &my_pcol, &izero, &np_cols);
+#if defined(SINGLE_REAL) || defined(DOUBLE_REAL)
+    int lwork = MAX(1 + 6 * m + 2 * np0 * nq0,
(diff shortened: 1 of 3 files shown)


cnegre (approved):

src/C-interface/distributed2d/bml_diagonalize_distributed2d_typed.c
```diff
@@ -237,6 +237,24 @@ void TYPED_FUNC(
     char jobz = 'V';
     char uplo = 'U';
     int info;
+    //LOG_INFO("lwork=%d\n",lwork);
+    lwork*=2; // increase lwork to work around ScaLapack possible bug
```
nicolasbock, line 241: Please fix the lniter issue here.

nicolasbock (changes requested):
Just a few linter issues, but otherwise looks good.

suemni (approved):

src/C-interface/distributed2d/bml_diagonalize_distributed2d_typed.c
```diff
@@ -237,6 +237,24 @@ void TYPED_FUNC(
...
+    lwork *= 2;                 // increase lwork to work around ScaLapack possible bug
+    REAL_T *work = bml_allocate_memory(lwork * sizeof(REAL_T));;
+
+    //// get lwork value from ScaLapack call
+    //// for verification
+    //lwork=-1;
+    //SYEVD(&jobz, &uplo, &m, atmp, &ione, &ione, desc, ev,
+    //      pzmat, &ione, &ione, desc, work, &lwork,
+#if defined(SINGLE_COMPLEX) || defined(DOUBLE_COMPLEX)
+    //      rwork, &lrwork,
```
nicolasbock, line 250: Is this comment inside an `ifdef` on purpose? If the commented lines are meant as documentation I would also comment out the `ifdefs` to make that more clear.

nicolasbock (changes requested):

src/C-interface/distributed2d/bml_diagonalize_distributed2d_typed.c
jeanlucf22: ifdef commented out now

nicolasbock: Thanks!

nicolasbock (approved):

PR #562, opened 2021-11-23
Element-wise matrix multiplication and corresponding tests are added

ares201005:
The new response calculation API in qmd/progress for the kernel in LATTE requires element-wise matrix multiply, which is implemented in this PR.

### DIFF
--- a/tests/C-tests/element_multiply_matrix_typed.c
+++ b/tests/C-tests/element_multiply_matrix_typed.c
@@ -1,0 +1,226 @@
+#include "bml.h"
+#include "../macros.h"
+#include "../typed.h"
+#include "ellblock/bml_allocate_ellblock.h"
+#include "bml_utilities.h"
+
+#include <complex.h>
+#include <math.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+#ifdef _OPENMP
+#include<omp.h>
+#else
+#include<time.h>
+#endif
+
+static void TYPED_FUNC(
+    ref_element_multiply) (
+    const int N,
+    const REAL_T * A,
+    const REAL_T * B,
+    REAL_T * C,
+    const double alpha,
+    const double beta,
+    const double threshold)
+{
+    for (int i = 0; i < N; i++)
+    {
+        for (int j = 0; j < N; j++)
+        {
+            C[ROWMAJOR(i, j, N, N)] *= beta;
+            C[ROWMAJOR(i, j, N, N)] +=
+                alpha * A[ROWMAJOR(i, j, N, N)] * B[ROWMAJOR(i, j, N, N)];
+        }
+    }
+}
+
+#if defined(SINGLE_REAL) || defined(SINGLE_COMPLEX)
+#define ABS_TOL 2e-6
+#else
+#define ABS_TOL 1e-12
+#endif
+
+static int TYPED_FUNC(
+    compare_matrix) (
+    const int N,
+    const bml_matrix_precision_t matrix_precision,
+    REAL_T * A,
+    REAL_T * B)
+{
+    int max_row = MIN(N, PRINT_THRESHOLD);
+    int max_col = MIN(N, PRINT_THRESHOLD);
+
+    LOG_INFO("Compare matrices\n");
+    for (int i = 0; i < N * N; i++)
+    {
+        if (ABS(A[i] - B[i]) > ABS_TOL)
+        {
+            LOG_INFO("First matrix\n");
+            bml_print_dense_matrix(N, matrix_precision, dense_row_major, A, 0,
+                                   max_row, 0, max_col);
+            LOG_INFO("Second matrix\n");
+            bml_print_dense_matrix(N, matrix_precision, dense_row_major, B, 0,
+                                   max_row, 0, max_col);
+            LOG_INFO("element %d outside %1.2e\n", i, ABS_TOL);
+#if defined(SINGLE_COMPLEX) || defined(DOUBLE_COMPLEX)
+            LOG_INFO("A[%d] = %e+%ei\n", i, creal(A[i]), cimag(A[i]));
+            LOG_INFO("B[%d] = %e+%ei\n", i, creal(B[i]), cimag(B[i]));
+            LOG_INFO("abs(A[%d]-B[%d]) = %e\n", i, i, ABS(A[i] - B[i]));
+#else
+            LOG_INFO("A[%d] = %e\n", i, A[i]);
+            LOG_INFO("B[%d] = %e\n", i, B[i]);
+            LOG_INFO("abs(A[%d]-B[%d]) = %e\n", i, i, ABS(A[i] - B[i]));
+#endif
+            return 1;
+        }
+    }
+    return 0;
+}
+
+static void TYPED_FUNC(
+    setup_bsizes) (
+    const int N,
+    const int M)
+{
+    int bsizes[N];
+    int count = 0;
+    int nb = 0;
+    for (int i = 0; i < N; i++)
+    {
+        int bsize = (3 + i) % 6 + 1;
+        count += bsize;
+        if (count > N)
+        {
+            bsize -= (count - N);
+            count = N;
+        }
+        bsizes[i] = bsize;
+        if (count == N)
+        {
+            nb = i + 1;
+            break;
+        }
+    }
+    bml_set_block_sizes(bsizes, nb, nb * M / N);
+    for (int i = 0; i < nb; i++)
+    {
+        printf("bsizes[%d] = %d\n", i, bsizes[i]);
+    }
+
+}
+
+int TYPED_FUNC(
+    test_element_multiply) (
+    const int N,
+    const bml_matrix_type_t matrix_type,
+    const bml_matrix_precision_t matrix_precision,
+    const int M)
+{
+    // set block sizes for ellblock tests
+    // (unused by other tests)
+#ifdef DO_MPI
+    if (bml_getNRanks() > 1)
+    {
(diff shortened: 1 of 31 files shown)


suemni (approved):

ares201005:
Thanks. I will make the changes accordingly. 

PS: The code is correct for sparse matrices. The Fortran tests include both dense and sparse matrices. I will add C tests as well. 

> A few comments
> 
>     1. I would make it a separate test, not in the same file as regular matrix multiply testing. I would even put the implementation in different files, not the same used for regular multiplications
> 
>     2. I would add a C test as we do for other functionalities
> 
>     3. While I realize that it is not specific to that PR, the current test involve dense matrices, a special "case". Does it give you confidence the code is correct for sparse matrices?

jeanlucf22:
> Thanks. I will make the changes accordingly.
> 
> PS: The code is correct for sparse matrices. The Fortran tests include both dense and sparse matrices. I will add C tests as well.
> 
> > A few comments
> > ```
> > 1. I would make it a separate test, not in the same file as regular matrix multiply testing. I would even put the implementation in different files, not the same used for regular multiplications
> > 
> > 2. I would add a C test as we do for other functionalities
> > 
> > 3. While I realize that it is not specific to that PR, the current test involve dense matrices, a special "case". Does it give you confidence the code is correct for sparse matrices?
> > ```

I believe we are testing for sparse formats, but with 100% filled entries

ares201005:
> > Thanks. I will make the changes accordingly.
> > PS: The code is correct for sparse matrices. The Fortran tests include both dense and sparse matrices. I will add C tests as well.
> > > A few comments
> > > ```
> > > 1. I would make it a separate test, not in the same file as regular matrix multiply testing. I would even put the implementation in different files, not the same used for regular multiplications
> > > 
> > > 2. I would add a C test as we do for other functionalities
> > > 
> > > 3. While I realize that it is not specific to that PR, the current test involve dense matrices, a special "case". Does it give you confidence the code is correct for sparse matrices?
> > > ```
> 
> I believe we are testing for sparse formats, but with 100% filled entries

I see. Thanks.

nicolasbock (changes requested):
As @jeanlucf22 suggested, please add a separate test for sparse matrices and the new multiply. I understand that you tested the code and are confident that it works with sparse matrices. However, lhe point of unit tests is to detect regressions. The code you have now might work now but future code changes might break it.

Since the library is written in C the unit test needs to be written in C to test it without also going through the Fortran interface. That just adds an uneccessary indirection.

The Fortran test is necessary for testing the Fortran intereface.

Please update the comments to make it clear what you mean with elemenwise multiplication.

jeanlucf22 (commented):
@ares201005  There is a lot of commented out code that probably should be removed, unless it serves a purpose. Otherwise, I would approve this PR.

ares201005:
> @ares201005 There is a lot of commented out code that probably should be removed, unless it serves a purpose. Otherwise, I would approve this PR.

forgot to remove it, sorry. I will make a new commit.

src/C-interface/ellblock/bml_element_multiply_ellblock_typed.c
```diff
@@ -0,0 +1,332 @@
+#include "../../internal-blas/bml_gemm.h"
```
jeanlucf22, line 1: @ares201005 It looks like this function is never called by the higher level interface functions, right? Is it because it is not working? I'd prefer a PR without this file then. If it is working, then the test for it should be activated

ares201005: right, I need to remove it in this PR. This are still some issues to be fixed. The ellblock and distributed2d will be in the next PR.

jmohdyusof:
Are your local tests also using 4 ranks?

ares201005:
> Are your local tests also using 4 ranks?

yes, I run the ./script/ci-MPI-single-script.sh for the test, and confirmed it's 4 rank from the output file.

jmohdyusof:
OK, there are some things in the test code (bml_free_memory called from rank 0, no barrier) so maybe the export_to_dense is happening before the free completes.

ares201005:
> OK, there are some things in the test code (bml_free_memory called from rank 0, no barrier) so maybe the export_to_dense is happening before the free completes.

in which test code? Maybe I can add the fix in PR.

jmohdyusof:
tests/C-tests/normalize_matrix_typed.c 
I assume this is what is being run that generates the error: line 71 calls normalize, 72 frees memory (only from rank 0), etc

ares201005:
> tests/C-tests/normalize_matrix_typed.c
> I assume this is what is being run that generates the error: line 71 calls normalize, 72 frees memory (only from rank 0), etc

ok, I will fix it and include it in this PR.

jmohdyusof:
I can't guarantee this is the issue, it just looked a little weird to me. Some 'free memory' calls are called by all ranks, etc.

jeanlucf22:
The call to "free" from PE0 only is correct: in the distributed case, bml_export_to_dense return a dense matrix on PE0 only (that is the complete matrix made of all its distributed parts)

jeanlucf22:
I can reproduce the issue on a computer at ORNL. Looking into it.

jeanlucf22:
PR #569 should fix that issue

jmohdyusof:
The PR makes sense to me, but I am still curious: do we really need to free B_dense on rank0 before writing to it with export_to_dense?

jeanlucf22:
> 
If you don't, you would probably have a memory leak as you would assign a new memory allocation to that pointer (bml_export_to_dense allocates memory and return a pointer to that newly allocated memory)

nicolasbock:
@jmohdyusof You could remove the `free` and see whether `valgrind` starts complaining.

tests/C-tests/element_multiply_matrix_typed.c
```diff
@@ -0,0 +1,226 @@
...
+            LOG_INFO("abs(A[%d]-B[%d]) = %e\n", i, i, ABS(A[i] - B[i]));
+#endif
+            return 1;
+        }
+    }
+    return 0;
+}
+
+void TYPED_FUNC(
+    setup_bsizes2) (
```
jeanlucf22, line 83: what is this function for?

nicolasbock:
@ares201005 Please rebase this change on top of `master`. I just merged lanl/bml#569.

tests/C-tests/element_multiply_matrix_typed.c
ares201005: set block size, copied from setup_bsize in multiply_matrix_type.c.

ares201005:
> @ares201005 Please rebase this change on top of `master`. I just merged #569.

sure

jmohdyusof:
> > 
> 
> If you don't, you would probably have a memory leak as you would assign a new memory allocation to that pointer (bml_export_to_dense allocates memory and return a pointer to that newly allocated memory)

I guess it seems fragile to me, not having the export function check, or being able to free (or reuse) the previous allocated memory.

tests/C-tests/element_multiply_matrix_typed.c
ares201005: will be removed in rebasing this commit.

nicolasbock:
Hmm, there seems to be some way with `glibc` using [malloc_usable_size](https://man7.org/linux/man-pages/man3/malloc_usable_size.3.html).

If we were to use that function we will have to change the API to the export functions. Right now they return a pointer to the dense matrix. The API would have to take that pointer as an argument.

jmohdyusof:
It is probably too much work? In that case you replace the need to free memory in the calling routine with the need to allocate memory in the calling routine, so it is just different, but not really any less prone to error?

nicolasbock:
When I wrote this API I modeled it after `malloc()`. Given the potential pitfalls I probably shouldn't have.

nicolasbock:
I would be ok with changing the API to make memory handling more robust though.

jmohdyusof:
I think this touches enough code (we use export_to_dense pretty much everywhere) that I would defer it until a major rewrite? If we ever could move to C++ templates and not have to write everything 4 times that would be a good time :)

nicolasbock:
C++ templates would be awesome 😄 

We should think about whether we want to eventually go there. It should be possible to gradually refactor the code to introduce templates, no?

ares201005:
now all tests are passed. Please let me know if further changes are required.

jmohdyusof:
> C++ templates would be awesome 😄
> 
> We should think about whether we want to eventually go there. It should be possible to gradually refactor the code to introduce templates, no?

There were reasons that C++ was deemed not appropriate. I don't recall them now, but I don't think it will help to rehash that again.

nicolasbock:
From what I remember a concern was data interoperability between C and Fortran (and maybe Python) with the library. C++ compilers have this convoluted name mangling which makes interfacing potentially tricky.

CMake has some functionality to detect whatever naming convention de jour the compiler is using though and the above concern could be mitigated that way.

jmohdyusof:
If that is the issue, maybe it makes sense to do a small test code? Interoperability is certainly a problem already, but being able to cut the code volume down significantly seems like it might be worthwhile?

nicolasbock:
Hmm, yes, that's a good point. I'll create an issue so we can track this effort.

nicolasbock:
I have created issue lanl/bml#570 to track a potential transition to `C++`.

jeanlucf22 (approved):

nicolasbock (approved):
Thanks @ares201005 !

ares201005:
Thank you!

PR #567, opened 2021-12-03
Fix permission issue with stale workflow

nicolasbock:
The permissions might have to be set explicitly in the workflow [1].

[1] https://github.com/actions/stale#recommended-permissions


nicolasbock:
The workflow does not have sufficient permission to change an issue [1].

[1] https://github.com/lanl/bml/runs/4410535117?check_suite_focus=true#step:2:409

PR #575, opened 2022-01-26
Add RTD badge

nicolasbock:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,6 +5,7 @@
     - Pull Requests
     - CI
     - Conda
+    - Documentation
   * - .. image:: https://img.shields.io/github/issues/lanl/bml
         :alt: GitHub issues
         :target: https://github.com/lanl/bml/issues>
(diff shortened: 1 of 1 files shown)


suemni (approved):

jmohdyusof:
Doesn't it make sense to change the docs branch to be the default for RTD for now, so that you don't have to merge with master every time?

nicolasbock:
RTD is building from `master`. Currently we don't have a separate `docs` branch.

I would prefer if we left the documentation in the `master` branch so that we don't have to sync branches to keep the documentation up to date.

nicolasbock:
@jmohdyusof It turns out that RTD is using a really old `sphinx` version (`sphinx-1.8.6`) which I didn't catch when testing the documentation build locally. I have submitted lanl/bml#576 to address this.

jmohdyusof:
> RTD is building from `master`. Currently we don't have a separate `docs` branch.
> 
> I would prefer if we left the documentation in the `master` branch so that we don't have to sync branches to keep the documentation up to date.

I was suggesting to make a docs branch and set RTD to use that temporarily so that you (and I?) can edit it on the fly to make docs work, then we can merge to master and set RTD to use master going forward. Otherwise every edit requires a pull request.

nicolasbock:
Ok I see. I created a `docs` branch and set RTD to build from it.

jmohdyusof:
Cool, let me know if you want me to help with hacking on that.

PR #602, opened 2022-03-23
Cusparse dev

oseikuffuor1:
This is a draft PR to show the cusparse integration within the ellpack format. A few additional details need to be addressed before a thorough review and merge into master:
1. Current cusparse support is for the matrix multiplication routine only: C = alpha*A*B + beta*C
2. No support for post dropping of small entries has been implemented yet. We can potentially do this with a cusparse routine, but it is not clear if that will be a win.
3. The current implementation converts between ellpack and and cusparse csr formats before and after the matrix multiplication. Could we make the cusparse data structure persist for multiple operations before call to update back to ellpack format? For example, SP2 using cusparse format? Are all the required operations feasible with cusparse?
4. Current implementation assumes sorted column index arrays (as required by cusparse). Is this assumption always true for ellpack?


jeanlucf22:
Looks like this PR is composed of two different components: i) coupling to cusparse ii) adapt to offload with XL compiler. I would prefer if this was split into two PRs.

jmohdyusof:
The standard ellpack functions (Gustavson algorithms) do not guarantee the result is sorted. The merge-based algorithms do expect, and retain sorting. 
It appears there are cusparse versions of multiply_with_threshold, add, trace so I think there should be no problem doing a full SP2 loop in cusparse format?

oseikuffuor1:
> 
@jmohdyusof is there a flag to turn on the merge-based option?

jmohdyusof:
AFAICT all the merge-based routines have been purged from the repo

jeanlucf22:
@oseikuffuor1 Can we close this one?

oseikuffuor1:
@jeanlucf22 yes, I will take care of that.

PR #607, opened 2022-04-21
Add simple test to verify OpenMP offload

jeanlucf22:
Verify simple OpenMP loop running on device,
to check OpenMP offload working

### DIFF
--- a/tests/C-tests/CMakeLists.txt
+++ b/tests/C-tests/CMakeLists.txt
@@ -1,5 +1,16 @@
 include_directories(${PROJECT_SOURCE_DIR}/src/C-interface)
 
+if(BML_OMP_OFFLOAD)
+  add_executable(test-openmp_offload test_openmp_offload.c)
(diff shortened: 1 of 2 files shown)


jeanlucf22:
> I am not sure this is doing what you think it is. From the OpenMP 5.2 spec document:
> 
> 18.7.6 omp_is_initial_device Summary The omp_is_initial_device routine returns true if the current task is executing on the host device; otherwise, it returns false.
> 
> I assume we want the code to be executing on the target device if offload is true?

0 = false. So if running on GPU, value will be 0 and test will pass

jmohdyusof:
OK, I guess I was confused because for C usually exiting with 0 = success. So long as it is used correctly, I guess it is fine

suemni (approved):

nicolasbock (approved):

nicolasbock:
I cannot test whether this works but the code looks good.

PR #625, opened 2022-06-20
Strengthen test get_set_diagonal

jeanlucf22:


jmohdyusof:
As always I would advocate some sort of label for the output, rather than just printing yet another matrix with no explanation of what it is.

jeanlucf22:
> 

No, LOG_INFO does not know anything about MPI

jmohdyusof:
Do we have a macro that only prints on rank 0, even better if it automatically disables itself if it is called from an offload region. Then I wouldn't get tripped up by putting bml_print_bml_matrix inside a guard region..

nicolasbock (approved):

jeanlucf22:
can we merge this one? thanks

PR #640, opened 2022-07-08
Lessen cmake requirements for non-GPU build

jeanlucf22:


CMakeLists.txt
```diff
@@ -213,12 +217,16 @@ if(BML_MPI)
   add_definitions(-DDO_MPI)
 endif()
 
-include(FindCUDAToolkit)
-if(CUDAToolkit_FOUND)
-  message("Found CUDAToolkit: ${CUDAToolkit_TARGET_DIR}")
-endif()  
+if(BML_CUDA OR BML_CUSOLVER OR BML_CUSPARSE)
```
mewall, line 220: What about BML_MAGMA?

CMakeLists.txt
```diff
@@ -213,12 +217,16 @@ if(BML_MPI)
...
+  include(FindCUDAToolkit)
+  if(CUDAToolkit_FOUND)
+    message("Found CUDAToolkit: ${CUDAToolkit_TARGET_DIR}")
+  endif()
+endif()
 
-find_package(hip QUIET)
+if(BML_ROCSOLVER OR BML_ROCSPARSE)
```
mewall, line 227: What about BML_MAGMA?

mewall, line 227: I think we can leave out this if() safely, it will just fail quietly if hip isn't found.

CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
-cmake_minimum_required(VERSION 3.17)
+if(BML_CUDA OR BML_CUSOLVER OR BML_CUSPARSE)
```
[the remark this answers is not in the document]
jeanlucf22: When linking with magma alone, magma.pc already contains the information about the needed cuda libraries and no additional library is needed

CMakeLists.txt
mewall: It looks like BML_MAGMA might be needed here, the following happens when building against MAGMA on crusher:

/sw/crusher/spack-envs/base/opt/cray-sles15-zen3/gcc-11.2.0/magma-2.6.1-4ns6vbbhaeosp36tm4jvl3e3ohxxrzpb/include/magma_types.h:148:14: fatal error: hipblas.h: No such file or directory
  148 |     #include <hipblas.h>

jeanlucf22: Ok, looks like magma is installed differently on crusher. Added BML_MAGMA flag

mewall: Thanks, that worked.

suemni (approved):

jeanlucf22:
Ready for review

mewall (approved):

nicolasbock (approved):

PR #649, opened 2022-08-04
Fix no-complex build for bml_elemental code

jeanlucf22:


suemni (approved):

src/C-interface/bml_elemental.c
```diff
@@ -77,7 +79,7 @@ bml_get_element_double_real(
  * \param A The bml matrix
  * \return The matrix element
  */
-float complex
+float _Complex
```
nicolasbock, line 82: Why change that to `_Complex`? They are the same but we are using `complex` everywhere else in the library.

jeanlucf22: We are using _Complex in https://github.com/lanl/bml/blob/master/src/typed.h

nicolasbock: Ah true, I missed that. But we are also using `complex` here

https://github.com/lanl/bml/blob/ad26f0bc71f87dcb26ccccfbf04e3a56d382e70d/src/C-interface/ellpack/bml_elemental_ellpack.c#L82-L86

for example.

Let's merge this PR then as it is and then follow up with a clean up so that the use of `comple` / `_Complex` is consistent throughout.

While `_Complex` is equivalent to `complex` the former is never mentioned in any API documentation. For example:

```
CABS(3)

NAME
       cabs, cabsf, cabsl - absolute value of a complex number

SYNOPSIS
       #include <complex.h>

       double cabs(double complex z);
       float cabsf(float complex z);
       long double cabsl(long double complex z);

       Link with -lm.
```

Also, symbols with preceding `_` are typically interpreted as `internal` symbols not meant for external use.

To make a long story short, I would prefer the consistent use of `complex` :smile:

nicolasbock (approved):

PR #662, opened 2022-09-28
Updated manual

cnegre:
Updating manual


suemni (approved):

jeanlucf22 (changes requested):
Why is this pdf in the repo? Isn't it a "generated" file? As such, I would keep only the source files in the repo, not the generated pdf.

nicolasbock:
We should provide this pdf as an artifact of a release. @cnegre Can you remove the pdf from this PR? I will follow up with another one that publishes the pdf for a release.

nicolasbock (changes requested):
Please remove the `pdf` file from the repository in this PR @cnegre

PR #669, opened 2022-11-15
Fix functions/values used for multiply x2 test

jeanlucf22:


jmohdyusof:
Should like 37 and 39 use the same format? I have no opinion on which is preferred (but happy to learn)

jeanlucf22:
> Should like 37 and 39 use the same format? I have no opinion on which is preferred (but happy to learn)

The 'f' is for float only, not double (if I understand it right)

jmohdyusof:
So is 2e-6 interpreted as a double?

jeanlucf22:
> 

I believe so, that is why you do not see a warning for that one

jmohdyusof:
This seems to be correct. I would prefer something like 2e-6f (but I am not sure it is allowed) just because I dislike having to count zeros

suemni (approved):

nicolasbock:
With `gcc-12.2.0` I get the following:

```C
#include <stdlib.h>
#include <stdio.h>

int main() {
    printf("sizeof() = %lu\n", sizeof(1e-6f));
}
```

```console
gcc test.c && ./a.out 
sizeof() = 4
```

So it is possible to write the literal in `e` notation (for @jmohdyusof :smile:) and also force it to be `float`.

nicolasbock:
I forgot to add that I would also prefer the 'e' notation. Counting zeros is error prone.

jmohdyusof (approved):
Approved, will test on Intel and Darwin

nicolasbock (approved):

PR #676, opened 2022-12-01
Fix tolerance in matrix norm test

jeanlucf22:
Scale tolerance with number of non-zero.
Now required for distribuide case which uses bigger matrices.


nicolasbock:
CI failures are not related to PR. I will re-run those tests once all tests are done.

suemni (approved):

nicolasbock (approved):

PR #709, opened 2023-04-12
Pull devcontainer from Docker Hub

nicolasbock:
Which is much faster than building a fresh container image every time.


nicolasbock:
A note: The idea here is that anyone can use GitHub's Codespace and will get a preconfigured container that includes all of the necessary compilers and tools just as we use in CI. No local installs necessary.

PR #727, opened 2023-08-22
Updated tranpose operations to support non-symmetric matrix structure.

oseikuffuor1:
* Modified ellpack and ellblock to correctly support arbitrary structure.
* Updated csr transpose algo. to be consistent with new ellpack algo.


jeanlucf22 (commented):
Why keep the old code? Removing it may also help in seeing the differences between the old and the new

jeanlucf22:
> > Why keep the old code? Removing it may also help in seeing the differences between the old and the new
> 
> @jeanlucf22 that was unintended for csr and ellblock. I will remove those. For ellpack, both are actually new code. I left comments at the beginning of the code to explain the difference between the two. The first is likely better for matrices with non symmetric structure, and the second is likely better for matrices with symmetric structure. For general matrices, I think either option should be fine. I left that in there so we can discuss it at our meeting tomorrow.

@oseikuffuor1 Looks good once you fix the formatting

oseikuffuor1:
> > > Why keep the old code? Removing it may also help in seeing the differences between the old and the new
> > 
> > 
> > @jeanlucf22 that was unintended for csr and ellblock. I will remove those. For ellpack, both are actually new code. I left comments at the beginning of the code to explain the difference between the two. The first is likely better for matrices with non symmetric structure, and the second is likely better for matrices with symmetric structure. For general matrices, I think either option should be fine. I left that in there so we can discuss it at our meeting tomorrow.
> 
> @oseikuffuor1 Looks good once you fix the formatting

@jeanlucf22, all fixed now. Thanks.

jeanlucf22 (approved):

PR #729, opened 2023-08-23
Bug fixes for norm calculation.

oseikuffuor1:
* Fixed various bugs in norm calculation that implicitly assumed dense storage for sparse matrices.
* This fix should enable correct norm calculations on arbitrary sparse matrices.

### DIFF
--- a/src/C-interface/ellpack/bml_norm_ellpack_typed.c
+++ b/src/C-interface/ellpack/bml_norm_ellpack_typed.c
@@ -190,38 +190,27 @@ double TYPED_FUNC(
         for (int jp = 0; jp < A_nnz[i]; jp++)
         {
             int k = A_index[ROWMAJOR(i, jp, A_N, A_M)];
-            if (ix[k] == 0)
-            {
-                y[k] = 0.0;
-                ix[k] = i + 1;
-                jjb[l] = k;
-                l++;
-            }
-            y[k] += alpha_ * A_value[ROWMAJOR(i, jp, A_N, A_M)];
+            ix[k] = i + 1;
+            y[k] = alpha_ * A_value[ROWMAJOR(i, jp, A_N, A_M)];
         }
 
         for (int jp = 0; jp < B_nnz[i]; jp++)
         {
             int k = B_index[ROWMAJOR(i, jp, B_N, B_M)];
-            if (ix[k] == 0)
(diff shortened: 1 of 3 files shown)


suemni (approved):

src/C-interface/ellpack/bml_norm_ellpack_typed.c
```diff
@@ -190,38 +190,27 @@ double TYPED_FUNC(
...
         for (int jp = 0; jp < l; jp++)
         {
             if (ABS(y[jjb[jp]]) > threshold)
                 sum += y[jjb[jp]];      //* y[jjb[jp]];
-
-            ix[jjb[jp]] = 0;
-            y[jjb[jp]] = 0.0;
-            jjb[jp] = 0;
         }
+        memset(ix, 0, A_N * sizeof(int));
```
jeanlucf22, line 213: I guess this memset could be done at the beginning of the loop and not be IBM specific?

oseikuffuor1: @jeanlucf22, sure, but perhaps a loop to reset the entries is more appropriate here?

jeanlucf22: why a loop? 
BTW, the array sizes could be A_M instead on A_N.
Also: why are we setting a value of i+1, when it just needs to be something else than 0? It is correct but can be confusing...

oseikuffuor1: A loop so we don't have to worry about the IBM guards?
Also:
True, that was the original line, but it is really only a marker so using 0/1 should be fine.

jeanlucf22: I meant move memset(ix, 0, A_N * sizeof(int)); after the IBM guards, so that it is called all the time. The one at the end becomes unnecessary then.

oseikuffuor1: Yeah I got what you meant later :-). I am making the changes now.

oseikuffuor1: @jeanlucf22, changes applied

jeanlucf22 (approved):

PR #732, opened 2023-08-30
Reorganize "domain" code

jeanlucf22:


suemni (approved):

jeanlucf22:
> @cnegre @ares201005 Could you check if I did not break anything with those changes? We have no test for the domain struct in BML, but it seems to be used in PROGRESS

@cnegre @ares201005 : any comment?

jeanlucf22:
> 

@ares201005 Is there a need to add a new test in PROGRESS to make sure we don't break that functionality?

PR #733, opened 2023-09-12
Separate real offload diagonalize routines.

jmohdyusof:
Still seeing build issues.

### DIFF
--- a/src/C-interface/dense/bml_diagonalize_dense.c
+++ b/src/C-interface/dense/bml_diagonalize_dense.c
@@ -21,7 +21,14 @@
 #include <rocsolver.h>
 #endif
 #endif
+
+#ifdef MKL_GPU
+#include "stdio.h"
+#include "mkl.h"
+#include "mkl_omp_offload.h"
+#else
 #include "../lapack.h"
+#endif
 
 #include <string.h>
 #include <complex.h>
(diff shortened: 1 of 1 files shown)


src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -85,6 +92,135 @@ bml_diagonalize_dense_magma_single_real(
...
+    float *typed_eigenvalues = (float *) eigenvalues;
+
+    float *evecs = (float *) malloc(A->N * A->N * sizeof(float));
+    float *evals = (float *) malloc(A->N * sizeof(float));
+
+#pragma omp target enter data map(alloc:evecs[0:N*N])
+#pragma omp target enter data map(alloc:evals[0:N])
+// pull from GPU
+#pragma omp target update from(A_matrix[0:N*N])
+    printf("Checking A matrix values \n");
```
jeanlucf22, line 119: These prints should be commented put for optimized built

src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -85,6 +92,135 @@ bml_diagonalize_dense_magma_single_real(
...
+#pragma omp target teams distribute parallel for
+    for (int i = 0; i < N; i++)
+    {
+        evals[i] = 1.0;
+    }
+
+#pragma omp target update from(evecs[0:N*N])
+#pragma omp target update from(evals[0:N])
+    // Check values before syev
+    printf("Checking evecs input values \n");
```
jeanlucf22, line 143: These too

src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -85,6 +92,135 @@ bml_diagonalize_dense_magma_single_real(
...
+        printf("\n");
+    }
+    // copy A to evecs on GPU
+#pragma omp target teams distribute parallel for
+    for (int i = 0; i < N*N; i++)
+    {
+        evecs[i] = A_matrix[i];
+    }
+#pragma omp target teams distribute parallel for
+    for (int i = 0; i < N; i++)
```
jeanlucf22, line 135: Do evals need to be initialized?

src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -85,6 +92,135 @@ bml_diagonalize_dense_magma_single_real(
...
+    {
+        for (int j = 0; j < N; j++)
+        {
+            e_matrix[ROWMAJOR(i, j, N, N)] = evecs[COLMAJOR(i, j, N, N)];
+        }
+    }
+
+// push eigenvectors back from GPU
```
jeanlucf22, line 214: remove these comments

src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -326,6 +458,143 @@ bml_diagonalize_dense_magma_double_real(
...
+    }
+#endif // debug
+*/
+#ifdef BML_SYEVD
+    // Divide and conquer solver
+    const MKL_INT lwork = 1 + 6 * N + 2 * N * N;
+    double *work = malloc(lwork * sizeof(double));
+    const MKL_INT liwork = 3 + 5 * N;
+    MKL_INT *iwork = malloc(liwork * sizeof(MKL_INT));
+#pragma omp target enter data map(alloc:work[0:lwork])
```
jeanlucf22, line 534: Are these temporary arrays ever deallocated?

src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -85,6 +92,135 @@ bml_diagonalize_dense_magma_single_real(
...
+// need typed_eigenvalues on CPU
+#pragma omp target update from(evals[0:N])
+    for (int i = 0; i < N; i++)
+    {
+        typed_eigenvalues[i] = (float) evals[i];
+    }
+
+// leave eigenvectors on GPU
+    float *e_matrix = (float *) eigenvectors->matrix;
+#pragma omp target enter data map(alloc:e_matrix[0:N*N])
```
jeanlucf22, line 204: Isn't this array already allocated? We just need to copy transposed results into it...

jeanlucf22 (changes requested):

src/C-interface/dense/bml_diagonalize_dense.c
jmohdyusof: There is no gpu array for eigenvectors allocated yet, unless I missed it. This was the original confusing reuse of A_matrix in the original code

src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -326,6 +458,143 @@ bml_diagonalize_dense_magma_double_real(
...
+    double *evecs = (double *) calloc(A->N * A->N, sizeof(double));
+    double *evals = (double *) calloc(A->N, sizeof(double));
+
+#pragma omp target enter data map(alloc:evecs[0:N*N])
+#pragma omp target enter data map(alloc:evals[0:N])
+
+// pull from GPU
+#pragma omp target update from(A_matrix[0:N*N])
+    printf("Checking A matrix values \n");
+    for (int i = 0; i < N; i++)
```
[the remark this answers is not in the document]
jmohdyusof: Am I allowed to leave them inside ifdefs for debugging?

src/C-interface/dense/bml_diagonalize_dense.c
jeanlucf22: Two arrays are passed in as arguments: A and eigenvectors. They are both BML matrices, and thus should have been allocated both on cpu and gpu when created. A contains the matrix (input), eigenvectors contains the computed eigenvectors (output). Since *syev* takes only one matrix argument as an input and output, we use "eigenvectors" for that, and copy A into eigenvectors before calling *syev*

src/C-interface/dense/bml_diagonalize_dense.c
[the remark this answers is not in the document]
jeanlucf22: I would say yes. I am aware that we still may need them...

src/C-interface/dense/bml_diagonalize_dense.c
jmohdyusof: Why isn't this done in the original code (evecs is allocated and used instead of eigenvectors)? I copied the original code as much as possible

jeanlucf22: It is true that in the magma code I introduced a third matrix- that maybe could be avoided. But the main issue here is that in the MKL_GPU code, eigenvectors should not be reallocated, but should be assigned to.

suemni (approved):

src/C-interface/dense/bml_diagonalize_dense.c
jeanlucf22: I think the only thing to change here is simply to remove the allocation for e_matrix on GPU. And I think the third matrix is indeed necessary to do the transpose without touching the originl A.

src/C-interface/dense/bml_diagonalize_dense.c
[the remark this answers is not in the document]
jeanlucf22: I still do not see debug guards around this one

src/C-interface/dense/bml_diagonalize_dense.c
```diff
@@ -488,6 +488,7 @@ bml_diagonalize_dense_gpu_double_real(
 #pragma omp target enter data map(alloc:evecs[0:N*N])
 #pragma omp target enter data map(alloc:evals[0:N])
 
+#ifdef MKL_GOU_DEBUG
```
jeanlucf22, line 491: typo

jeanlucf22 (approved):

PR #752, opened 2024-03-20
Bml whypre

oseikuffuor1:
This PR enables integrating hypre into BML to allow BML to access hypre's GPU capable sparse linear algebra routines, particularly, matrix multiplication and matrix addition.
* The approach should allow BML to use hypre for sparse linear algebra operations on NVIDIA, AMD and Intel GPUs. However, the current implementation targets NVIDIA GPUs. Minor edits/ additions are necessary to support AMD.
* Currently, hypre can be built in one precision at a time, so bml tests for multiple precisions will only pass for the precision for which hypre was built (if BML is built with hypre support).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,6 +229,7 @@ set(BML_ROCSPARSE FALSE CACHE BOOL "Whether to compile with rocSPARSE support")
 set(BML_CUDA FALSE CACHE BOOL "Whether to compile with CUDA support")
 set(BML_CUSOLVER FALSE CACHE BOOL "Whether to compile with cuSolver support")
 set(BML_MAGMA FALSE CACHE BOOL "Whether to use MAGMA library")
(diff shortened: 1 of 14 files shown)


nicolasbock (changes requested):
Could you

1. squash the commits
2. Clean up the commit message such that [[1]](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53)
a. The first line is at most 72 characters long
b. There is an empty line
c. Wrap the "body" of the message to 72 characters per line

[1] <https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53>

nicolasbock:
Yes, ideally :smile:

nicolasbock (approved):
Thanks!

oseikuffuor1:
Thanks @nicolasbock!

PR #754, opened 2024-06-28
Require BML_USE_POSIX_MEMALIGN to use posix_memalign()

mewall:
Using the HAVE_POSIX_MEMALIGN code path results in ~1.7x slowdown of an MD code which frequently allocates arrays.

Require a new macro BML_USE_POSIX_MEMALIGN to be defined, to use the code path. Define the macro by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -671,9 +671,14 @@ endif()
 
 add_definitions(-D_POSIX_C_SOURCE=200112L)
 
-check_function_exists(posix_memalign HAVE_POSIX_MEMALIGN)
-if(HAVE_POSIX_MEMALIGN)
-  add_definitions(-DHAVE_POSIX_MEMALIGN)
+option(BML_POSIX_MEMALIGN "Use posix_memalign() for allocations if available" ON)
+
+if(BML_POSIX_MEMALIGN)
+  check_function_exists(posix_memalign HAVE_POSIX_MEMALIGN)
+  if(HAVE_POSIX_MEMALIGN)
+    add_definitions(-DBML_USE_POSIX_MEMALIGN)
+    message(STATUS "Will use posix_memalign() for allocations")
+  endif()
 endif()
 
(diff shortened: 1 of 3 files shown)


suemni (approved):

nicolasbock:
Hi @mewall

Do you see this kind of slow-down on all machines you run the code on or just on a particular architecture / OS combination? If this is a general issue we might simply want to remove this code path all together.

mewall:
I don't know but it would be good to look into that. I'll try that next week. For now it's possible avoid the code path using compiler flag -UHAVE_POSIX_MEMALIGN.
Sent: Friday, June 28, 2024 8:15 PM
To: lanl/bml ***@***.***>
Cc: Wall, Michael E ***@***.***>; Mention ***@***.***>
Subject: [EXTERNAL] Re: [lanl/bml] Require BML_USE_POSIX_MEMALIGN to use posix_memalign() (PR #754)


Hi @mewall<https://urldefense.com/v3/__https://github.com/mewall__;!!Bt8fGhp8LhKGRg!D50ERlVT-COdnX_ko_oqxrlQihZZmfZkx1QkqwxQ8Nb8IAnW5bHqSmH6iVwVsodS5SeAQZNrQLnT62Cqhs74cpOM$>

Do you see this kind of slow-down on all machines you run the code on or just on a particular architecture / OS combination? If this is a general issue we might simply want to remove this code path all together.

nicolasbock (approved):

nicolasbock:
It would be good to enable the old behavior by default and optionally allow to disable it. This would make sure that the old behavior does not unexpectedly change.

nicolasbock:
It would be good to enable the old behavior by default and optionally allow to disable it. This would make sure that the old behavior does not unexpectedly change.

mewall:
@nicolasbock The behavior is now the same as before by default. Please merge if OK.

nicolasbock (approved):
Thanks!

PR #763, opened 2025-03-24
Build bml with magma (v2.9) and newer gcc compiler (v 12.4.0)

finkeljos:
Added include file <stdbool.h> before the <magma_v2.h> include. On my newer gcc compilers, C did not understand what the bool data type meant, which is something that magma uses. Thus, I could not build without concluding this .h file.

### DIFF
--- a/src/C-interface/bml_init.c
+++ b/src/C-interface/bml_init.c
@@ -2,7 +2,7 @@
 #include "bml_parallel.h"
 
 #ifdef BML_USE_MAGMA
-#include <stdbool.h>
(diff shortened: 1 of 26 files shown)


finkeljos:
@jeanlucf22 Not sure what this failure means?? Is it for a style check??

jeanlucf22:
> @jeanlucf22 Not sure what this failure means?? Is it for a style check??

yes

suemni (approved):

jeanlucf22 (approved):

